PR #2, opened 2017-07-05
Update docker

ztolley:
Updated the docker file to integrate with the DB and run migration on start.
Also updated Node to a recent version and added yarn, ready for adding webpacker.


Dockerfile
```diff
@@ -1,5 +1,64 @@
...
+    gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "$key" ; \
+  done
+
+RUN curl -SLO "https://nodejs.org/dist/v$NODE_VERSION/node-v$NODE_VERSION-linux-x64.tar.xz" \
+  && curl -SLO --compressed "https://nodejs.org/dist/v$NODE_VERSION/SHASUMS256.txt.asc" \
+  && gpg --batch --decrypt --output SHASUMS256.txt SHASUMS256.txt.asc \
+  && grep " node-v$NODE_VERSION-linux-x64.tar.xz\$" SHASUMS256.txt | sha256sum -c - \
+  && tar -xJf "node-v$NODE_VERSION-linux-x64.tar.xz" -C /usr/local --strip-components=1 \
+  && rm "node-v$NODE_VERSION-linux-x64.tar.xz" SHASUMS256.txt.asc SHASUMS256.txt \
+  && ln -s /usr/local/bin/node /usr/local/bin/nodejs
```
rgarner: This seems like a good way to patch a container with a new node. Is what we're saying here "let's patch the ruby container with a new node because these things should co-exist"? If so is this a sustainable approach for future stuff, or do we need to create our own hybrid node/ruby container and pull from that later? (Doesn't affect this pull request, which looks fine on this score as a starter for ten)

rgarner: (it looks like, for example, this is a [common need](https://hub.docker.com/r/starefossen/ruby-node/)

ztolley: We need a hybrid of both ruby and node so we can do one later unless there's a better one out there.

ztolley: The DB user is created by setting the environment vars when starting the postgres container.

rgarner: Ta!

rgarner: That one I linked to is Debian, is 0.0.1 rubies and 0.0.2 nodes behind semantically. It may be we have to 'maintain' one, at least minimally.

PR #3, opened 2017-07-05
Add base webpacker integration

ztolley:
Provides the most basic integration with webpacker to transpile js to allow the use of vue js to progressively enhance form components.


ztolley:
Just remembered I didn't add yarn install to the dockerfile and after this will have to do a Readme to remind people to say install ruby, node AND yarn and how packages are split across the two

Gemfile
```diff
@@ -5,31 +5,14 @@ git_source(:github) do |repo_name|
   "https://github.com/#{repo_name}.git"
 end
 
-
-# Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
```
rgarner: I agree totally with the deletion of this kind of mindless boilerplate.

Can we split stuff like this into separate "tidy up" PRs (or at least commits)? It's not directly associated with adding `webpacker`, so it adds a bit of extra mental parse time. By no means a serious example, but since we're half in "koan" stage, now's the time to start thinking of being kind to future us, who are just more frazzled versions of "now" us :)

PR #4, opened 2017-07-06
Multi page form session example

rgarner:
Don't merge, but the PR format is useful for discussion


app/controllers/questions_controller.rb
```diff
@@ -1,5 +1,37 @@
...
+    redirect_to(first_question_path)
+  end
+
+  def show
+    @template_name = params[:id]
+    @session_data = session_data
+  end
 
+  def create
+    session_data.reverse_merge!(question_params)
```
rgarner: @ztolley this is your `Object.assign` as was

app/models/question.rb
```diff
@@ -0,0 +1,14 @@
...
+  def self.first
+    ORDERED_KEYS.first
+  end
+
+  def self.next(current)
+    current_index = ORDERED_KEYS.index(current.to_sym)
+    return nil if current_index.nil?
+    ORDERED_KEYS[current_index.next]
+  end
+end
```
rgarner: Apologies for the bad line endings, I hadn't rebased onto your `.editorconfig` stuff. Will fix and `push -f`

ztolley: surely you're not using windows?

rgarner: No, for Unix you need a trailing LF, and my editor's been stopping short (hence the little red circle). Big no-no at GDS, largely because it buggered up the ending output of any `cat`-piped thing.

PR #6, opened 2017-07-06
Build on circle

rgarner:
Add one part `circle.yml` along with a ton of [workarounds](https://discuss.circleci.com/t/yarn-build-failure-old-version-finding-incompatible-packages/12259/8) and an empty `rspec-rails`, and we have a minimally-passing build.


circle.yml
```diff
@@ -0,0 +1,16 @@
+dependencies:
+  override:
+    - yarn
+  pre:
+    - rm -rf ~/.yarn
+    - npm install -g yarn
```
ztolley: Circle already has yarn installed.

I have used something like this before:

```
machine:
  timezone: Europe/London
  node:
    version: 8.1.0
dependencies:
  override:
    - yarn install && npm rebuild
test:
  override:
    - yarn circle:
        environment:
          MOCHA_FILE: $CIRCLE_TEST_REPORTS/test-results.xml
  post:
     - cp -R coverage $CIRCLE_ARTIFACTS
```

rgarner: It does. It's 21.x, and it gives the error detailed in the workarounds link above, namely:

```
The engine "node" is incompatible with this module. Expected version ">=4.3.0 <5.0.0 || >=5.10".
```

Using the workarounds in order made things work again.

rgarner: Your example seems also to have `yarn install` in it (but followed by `npm rebuild`)? Is it another way to skin the same cat? Don't know enough about `npm` and `yarn` to comment.

rgarner: Ok, I'm clearly too tired for this. Setting node's version on its own does it with yarn 0.21.3, so I've removed the reinstall of yarn as suggested.

PR #7, opened 2017-07-11
New base styles

ztolley:
Removed the styles from the 'old' Barnardos elements in preparation for creating styles based on new designs.

1st step is this PR which consists of removing the old styles and introducing base styles for:
* Headings
* Paragraphs
* Colours
* Header and content section


app/views/gallery/index.html.erb
```diff
@@ -0,0 +1,20 @@
+<h1 class="title">Titlas asd ad asd ada ade</h1>
+<h2 class="subtitle-large">Subtitle 1 a da asd ad adad asd asds</h2>
+<h3 class="subtitle-medium">Subtitle 2 asda dasdasd adad asd</h3>
+<h4 class="subtitle-small">Subtitle 3 asda sasdasdasdasdasd as dasdasd d</h4>
```
rgarner: Can we put something like "A large subtitle", "A medium subtitle" here?

rgarner:
We are busy seeing what the review tools do if one person doesn't approve. Don't panic, @ztolley :grin:

PR #9, opened 2017-07-12
Barnardos helper examples

rgarner:
Examples of how to make HTML helpers with pure Ruby. Use these to replace
nunjucks-generated markup brought over from `consent-form-helpers`:

* `#labelled_text_field_tag`
* `#vertical_radio_list`


spec/helpers/barnardos/form_helpers_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+      let(:name) { :participant_name }
+      let(:label) { 'What is the name of the research participant?' }
+
+      subject(:rendered) { helper.labelled_text_field_tag(name, label) }
+
+      it 'outputs an wrapper div with the class and id' do
+        expect(rendered).to have_tag('div', with: { class: 'textfield', id: 'participant_name-wrapper' })
+      end
+
+      it 'renders a classed bold label and input as sibling children of the div' do
```
ztolley: Can we test not only that it has the elements but also has the correct structure and order?

We'll need to add more tests in time to confirm things like picking up optional params like hint and error

rgarner: There are three tests:

```ruby
      it 'outputs an wrapper div with the class and id' do
        expect(rendered).to have_tag('div', with: { class: 'textfield', id: 'participant_name-wrapper' })
      end

      it 'renders a classed bold label and input as sibling children of the div' do
        expect(rendered).to have_tag('div.textfield > label.textfield__label.textfield__label--bold', text: label)
        expect(rendered).to have_tag('div.textfield > input.textfield__input')
        expect(rendered).to have_tag('label[for=participant_name]')
      end

      it 'names the input as given' do
        expect(rendered).to have_tag('input', with: { name: 'participant_name', type: 'text' })
      end
```

We *could* have one test which fails if *anything* is wrong, and this is something I've done in the past. Why am I not doing it here? Independent failure. If we wrote one test the `it` would have to read "Render a labelled input box inside a wrapper div that's all classed correctly`. You can do that, but if it fails you have to parse everything that *might* be wrong, and the most important thing in terms of a working site is that the input is named correctly (which is why it's got its own test – I'm less sure about the wrapper `div`).

Ordering is a harder problem in pure RSpec, there aren't any explicit before/after matchers for that kind of thing, so you need to fall back on `first-child`, `last-child`, `nth-child` in the CSS selectors (could add those in the middle test).

You can argue, probably quite successfully, that I've split these up badly. Do you fancy having a go at making it better?

You're right about the specs that need to follow. We'll need new `context`s for these.

ztolley: I think this is fine for now and I'll have a look at something that compares whole fragments and is html aware. I agree that I don't like tests that have a big fail and each test normally would focus on one thing.

Could also argue that this level of tests is perfect for unit like testing and layout should be done in acceptance tests that would be run against real browsers

rgarner:
@ztolley here's what that test would look like condensed into one:

```ruby
      it 'outputs an wrapper div with the class and id'\
         'containing a classed bold label and a named input' do
        expect(rendered).to have_tag('div', with: { class: 'textfield', id: 'participant_name-wrapper' }) do
          with_tag('div.textfield > label.textfield__label.textfield__label--bold:first-child[for=participant_name]', text: label)
          with_tag('div.textfield > input.textfield__input[name=participant_name][type=text]:last-child')
        end
      end
```

PR #10, opened 2017-07-12
Setup Linting tools for JS files.

ztolley:
Introduced eslint to keep styling of js consistent where possible. Uses rules defined by https://standardjs.com/ .

Can be used in an ide by installing an eslint module or from the command line and ci integration using npm run lint

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,2 @@
+public/
+manifest.js
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,4 @@
+{
+  "extends": "standard",
+  "parser": "babel-eslint"
+}
--- a/circle.yml
+++ b/circle.yml
@@ -11,4 +11,4 @@ machine:
     version: v8.1.3
 test:
   override:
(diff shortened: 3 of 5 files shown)


app/assets/config/manifest.js
```diff
@@ -1,2 +1,2 @@
-//= link_tree ../images
-//= link_directory ../stylesheets .css
+// = link_tree ../images
+// = link_directory ../stylesheets .css
```
rgarner: Any particular reason these have gained extra spaces?

rgarner: (These get processed by Sprockets. I'm not sure if the spaces would break that, but I guess us having webpacker means we aren't using Sprockets any more anyway. If that's the case is `manifest.js` still required?)

ztolley: we are still currently using rails built in sass compiler but not sure about how this bit works. the js stuff gets its own manifest created by webpacker.

ztolley: i will revert this file and exclude it from eslint

PR #12, opened 2017-07-13
Add new styles and behaviour for textfield

ztolley:
Created a basic textfield based on new designs.

* Updated textfield markup
* New css for textfield layout
* Added js to enhance textfield to add new focus on and off state


app/views/gallery/index.html.erb
```diff
@@ -18,3 +18,44 @@
...
+  <input
+    type='text'
+    class='textfield__input'
+    id="name"
+    name="name"
+    value="Jane Doe"
+    autoComplete="no"
+  />
+  <div id="name-error" class='textfield__error'>Here is an error</div>
+</div>
```
rgarner: Should we render these examples with the `Barnardos::ActionView::FormHelpers#labelled_text_field_tag` method itself?

app/javascript/lib/element-stuff.js
```diff
@@ -0,0 +1,47 @@
+/* eslint no-useless-escape: 0 */
+const regularExp1 = '(\\s|^)'
+const regularExp2 = '(\\s|$)'
```
rgarner: Can we call these something more intention-revealing?

app/javascript/lib/element-stuff.js
```diff
@@ -0,0 +1,47 @@
+/* eslint no-useless-escape: 0 */
```
rgarner: Is there something more descriptive we can call this file?

app/views/gallery/index.html.erb
ztolley: i was using this to test before that was merged and was getting the markup right first

app/javascript/lib/element-stuff.js
ztolley: maybe.

app/assets/stylesheets/barnardos/_colours.scss
```diff
@@ -7,8 +7,9 @@ $pink: #EC008C;
 $teal: #00A2AC;
 $purple: #781D7E;
 
-$error-color: #d0011b;
+$error-colour: #d0011b;
```
sl33p: 🇬🇧

package.json
```diff
@@ -29,11 +29,19 @@
   },
   "devDependencies": {
     "babel-eslint": "^7.2.3",
+    "chai": "^4.1.0",
+    "chai-as-promised": "^7.1.1",
     "eslint": "^4.2.0",
+    "jsdom": "^11.1.0",
+    "mocha": "^3.4.2",
```
rgarner: Question on running tests: I've got no `mocha` binary with a standard `yarn install`. What's the best way to ensure this shows up so I can run tests, locally and in Docker? I've exported `NODE_ENV=development` locally but `yarn install` still installs no `devDependencies`.

ztolley: By default Yarn installs dev dependencies. Try `npm run test:unit`

rgarner: Ah, ok, thanks. I assumed since we were using mocha there must be a mocha binary ... oops.

app/helpers/barnardos/action_view/form_helpers.rb
```diff
@@ -3,23 +3,29 @@ module ActionView
...
-      def labelled_text_field_tag(name, label, value: nil ,
+      def labelled_text_field_tag(name, label, value: nil , error: nil,
                                   text_options: {}, label_options: {})
-        content_tag :div, class: 'textfield', id: "#{name}-wrapper" do
+        className = 'textfield js-textfield'
+        if error
+          className += ' has-error'
+        end
+
+        content_tag :div, class: className, id: "#{name}-wrapper" do
```
rgarner: If you find it succinterer, you can use the post-script `if` form like you did on L28 to do something like:

```ruby
        content_tag :div, class: "textfield js-textfield #{' has-error' if error}", id: "#{name}-wrapper" do
```

That'd avoid the garbage collection for the `className` variable and the cost of the extra string concat. Rails views have a reputation for being slow, so whatever we can squeeze from helpers that we use multiple times helps.

ztolley: That's exactly what I wanted to do, just didn't know how

PR #16, opened 2017-07-22
Radio buttons

ztolley:
Created new styles for radio buttons based on updated designs.
CSS Names follow BEM
A radio group is all put within a fieldset, legends are required (though can be hidden) and hints are tied to a legend, though not options (an option should be obvious).
Updated tests to reflect new structure and different naming.
Currently this just supports vertical radio options, horizontal will come later when needed.


ztolley:
Just realised this doesn't check radio buttons where the field value == radio value.

spec/helpers/barnardos/radio_helpers_spec.rb
```diff
@@ -0,0 +1,159 @@
+require 'rails_helper'
+
+RSpec.describe Barnardos::ActionView::FormHelpers, :type => :helper do
```
ztolley: so outside of this PR show me how we can break it into multiple files.

app/helpers/barnardos/action_view/form_helpers.rb
```diff
@@ -26,28 +26,47 @@ def labelled_text_field_tag(name, label, value: nil , error: nil,
...
+              content_tag(:legend, class: "radio-group__legend #{legend_options[:class] if legend_options[:class]}") do
+                concat(legend)
+                concat(content_tag(:span, legend_options[:hint], class: 'radio-group__hint')) if legend_options[:hint]
               end
             )
+
+            # Render radio options
+            selection_list.each do |selection_item_value, selection_item_text|
+              id = "#{name}-#{selection_item_value}"
+              checked = value == selection_item_value
```
ztolley: its the whole Ruby style is weird thing.

PR #17, opened 2017-07-23
Checkbox

ztolley:
Created styled checkboxes and form helpers

### DIFF
--- a/spec/helpers/barnardos/checkbox_helper_spec.rb
+++ b/spec/helpers/barnardos/checkbox_helper_spec.rb
@@ -1,0 +1,145 @@
+require 'rails_helper'
+
+RSpec.describe Barnardos::ActionView::FormHelpers, :type => :helper do
+  include RSpecHtmlMatchers
+
+  describe '#checkbox_group_vertical' do
+    # <!-- Example HTML output -->
+    # <fieldset class="checkbox-group checkbox-group__vertical">
+    #   <legend class="checkbox-group__legend">
+    #     Some big legend
+    #     <span class="checkbox-group__hint">Some more help for the label</span>
+    #   </legend>
+    #
+    #   <div class="checkbox-group__choice">
+    #     <input class="checkbox-group__input" id="age-under12" type="checkbox" name="age" value="under12">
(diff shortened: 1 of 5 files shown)


spec/helpers/barnardos/checkbox_helper_spec.rb
```diff
@@ -0,0 +1,178 @@
+require 'rails_helper'
+
+RSpec.describe Barnardos::ActionView::FormHelpers, :type => :helper do
+  include RSpecHtmlMatchers
```
ztolley: so lets break them out of form helpers in a separate pr

spec/helpers/barnardos/checkbox_helper_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+
+    context 'a single field value is supplied' do
+      let(:values) { ['one'] }
+
+      it 'marks a single checkbox for a single value' do
+        expect(rendered).to have_tag('#age-one[checked="checked"]')
+      end
+      it 'does not check inputs that have not been selected' do
+        expect(rendered).not_to have_tag('#age-two[checked="checked"]')
+        expect(rendered).not_to have_tag('#age-three[checked="checked"]')
```
rgarner: This is a negative test, so it won't check that `#age-two` and `#age-three` are present but unchecked (in other words, it'll pass if either are absent). You could test with the selector instead:
```ruby
  expect(rendered).to have_tag('#age-two:not([checked])')
```

spec/helpers/barnardos/checkbox_helper_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+    context 'multiple field values are supplied' do
+      let(:values) { ['one', 'three'] }
+
+      it 'marks multiple checkboxes for multiple values' do
+        expect(rendered).to have_tag('#age-one[checked="checked"]', checked: true)
+        expect(rendered).to have_tag('#age-three[checked="checked"]', checked: true)
+      end
+
+      it 'does not check inputs that have not been selected' do
+        expect(rendered).not_to have_tag('#age-two[checked="checked"]', checked: true)
```
rgarner: Same as above

PR #19, opened 2017-07-24
Collect research session data via multi-step form

rgarner:
# [After entering my session details I can see a session review](https://trello.com/c/DDNtM82C/15-story-after-entering-my-session-details-i-can-see-a-session-review)

Before this PR: the project is pretty much a small tech demo. We:
* collect a couple of values to in-memory session in a homegrown multi-step wizard
* without validation, and
* don't put it anywhere

After this PR is merged, we:
* Will have thrown away our homegrown state management, replacing it with the [wicked](https://github.com/schneems/wicked) gem
* Will be collecting our values to a `ResearchSession` model, which builds up in each step
* Will have server-side validation of parameters
* Will have integrated the `radio_group_vertical` and `checkbox_group_vertical` helpers, our first Barnardo's elements
* Will be storing the state of each research session and whether it's finished or not (there's no URL to get back to a session finished or otherwise yet but there will be)

## Technical notes

* We now have half a Cucumber happy path acceptance test for building up the `ResearchSession`
* We've added our own validator for checkboxes and checking whether values are in the allowed range (and that there's at least one of them)
* We have not added separate DB-backed models for `RecordingMethods`, `Age`, `Methodologies` or `PaymentMethod`s. These are in-memory 'models' that are persisted as PostgreSQL arrays. This seemed faster and with fewer moving parts (read: migrations) when time-constrained.
* We **don't** have a session preview. This is in some regard half a story.
* We *will* end up with half-finished sessions that we periodically need to clean out


rgarner:
@ztolley define [releasable](https://github.com/barnardos/consent-form-builder-rails/pull/19/files#diff-6e7794c2e7657c77f69b1250c49e3774R1)?

PR #20, opened 2017-07-25
Use asset helpers to work with Heroku

ztolley:
Running Rails locally allows you to just put in the name of a font or use /assets/image to access assets but this breaks when deployed to Heroku.

Switched to helpers to load assets so they use the full hashed url on Heroku


rgarner:
Can you squash the fixup? Ta

rgarner:
Ok, squashed here, pushed, will merge when complete

PR #21, opened 2017-07-25
Session preview page

rgarner:
[After entering my session details I can see a session review](https://trello.com/c/DDNtM82C/15-story-after-entering-my-session-details-i-can-see-a-session-review)

## Job Story:

*When* I have filled in my research session details
*I want to* see a preview of the consent form content
*So I *can check that it is accurate and displays the way I want to.

## Acceptance Criteria

* Submitting session details leads to a session review page
* Session review page contains a preview of the information sheet and consent form

### DIFF
--- a/app/assets/stylesheets/app/session_preview.scss
+++ b/app/assets/stylesheets/app/session_preview.scss
@@ -1,0 +1,3 @@
+.session-preview-element__highlighted {
+  background-color: $orange;
+}
--- a/app/assets/stylesheets/app/session_preview.scss
+++ b/app/assets/stylesheets/app/session_preview.scss
@@ -1,3 +1,3 @@
-.session-preview-element__highlighted {
-  background-color: $orange;
+.highlight {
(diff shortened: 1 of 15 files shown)


app/assets/stylesheets/app/session_preview.scss
```diff
@@ -0,0 +1,3 @@
+.session-preview-element__highlighted {
```
ztolley: I would just call it highlight.

Also, what IS highlight? where did the need come from?

ztolley: highlight

If it is this then I'd be tempted to come up with a different name. So maybe we just call it 'highlight' and then have 

```
<div class="highlight">
  <h1 class="title">
  <p class="super-text">

```
Then css would be 

```
.highlight {
  background: $highlight-background-colour;
}
```

then in colours:

`$highlight-background-colour: #5b8126;`

PR #25, opened 2017-07-25
Created styles for error summary and clearer messages

ztolley:
* Styled error summary
* Put summary at top of screen above title
* Add link from summary item to field to allow navigation to field (fields currently missing this but will add later)

errors


rgarner:
Happy to have this in as-is (definite improvement) and iterate

PR #26, opened 2017-07-26
Display error for radio lists if provided

ztolley:
error
The radio control accepts a error parameter, which is an error by
convention. The control will then conditionally set the has-error class
on the wrapper and display the first error below the field.

### DIFF
--- a/app/assets/stylesheets/barnardos/_radio.scss
+++ b/app/assets/stylesheets/barnardos/_radio.scss
@@ -3,17 +3,34 @@ $inner-ring: 6px;
 $border-ring: $inner-ring + $ring-size;
 $outer-ring: $inner-ring + ($ring-size * 3);
 
+.radio-group {
+  margin: $gutter 0 0 (-$gutter / 2) ;
+  padding: $gutter / 2;
+}
+
+.radio-group.has-error {
+  background: $error-background;
+}
+
 .radio-group__legend {
   font-size: 23px;
   font-weight: normal;
   line-height: 24px;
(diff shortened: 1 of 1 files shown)


app/helpers/barnardos/action_view/form_helpers.rb
```diff
@@ -90,6 +91,12 @@ def radio_group_vertical(name, legend, selection_list, value = nil, error: nil,
                 end
               )
             end
+            concat(
+              content_tag(:div,
+                          "#{name.to_s.humanize} #{error[0]}",
+                          class: 'radio-group__error'
+              )
+            ) if error.present?
```
rgarner: If testing for `nil`, can just say `if error` here (any truthy value will do) – `present` is for when you need to test for nil or empty, in the case of a `String` that might be `""` or an array that might be `[]`

app/helpers/barnardos/action_view/form_helpers.rb
```diff
@@ -90,6 +91,14 @@ def radio_group_vertical(name, legend, selection_list, value = nil, error: nil,
                 end
               )
             end
+
+            concat(
+              content_tag(:div,
+                          "#{name.to_s.humanize} #{error[0]}",
```
rgarner: Should this just be a string at this stage? We have the singular `error` but we're passing an array.

ztolley: see checkbox comment

rgarner:
@ztolley could you rebase? Thanks

PR #27, opened 2017-07-26
Display error for checkbox lists if provided

ztolley:
methoderro

### DIFF
--- a/app/assets/stylesheets/barnardos/_checkbox.scss
+++ b/app/assets/stylesheets/barnardos/_checkbox.scss
@@ -3,17 +3,34 @@ $inner-ring: 6px;
 $border-ring: $inner-ring + $ring-size;
 $outer-ring: $inner-ring + ($ring-size * 3);
 
+.checkbox-group {
+  margin: $gutter 0 0 (-$gutter / 2) ;
+  padding: $gutter / 2;
+}
+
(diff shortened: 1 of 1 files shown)


app/helpers/barnardos/action_view/form_helpers.rb
```diff
@@ -118,6 +118,12 @@ def checkbox_group_vertical(name, legend, selection_list, values = [], error: ni
                 end
               )
             end
+            concat(
+              content_tag(:div,
+                          "#{name.to_s.humanize} #{error[0]}",
```
rgarner: The variable name says "error" but this is still an array. Can it just be a string?

rgarner:
'k, clearly I need to alter the validator, which is spitting out some human-unfriendly stuff

rgarner:
Am I right in saying this isn't referenced from either checkbox-using view at present?

app/helpers/barnardos/action_view/form_helpers.rb
ztolley: The validators generate an array of errors. It's design to work with data passed from the validator

ztolley:
Currently not of our pages put supply the error. I have a branch though you could look at that does for 2 screens

PR #28, opened 2017-07-26
Make bin/dockerstart.sh executable

rgarner:
Without this, docker-compose up won't work on non-Windows machines.
The Rails container will fail to start when attempting to fork a new
process.

e.g.

```
rails_1    | 2017/07/26 13:20:30 Error starting command: `./bin/dockerstart.sh` - fork/exec ./bin/dockerstart.sh: permission denied
```


rgarner:
We need to make sure CI is [running through Docker](https://trello.com/c/ssaUjtxg/18-make-sure-ci-is-building-using-docker) – that would have picked this up during the acceptance tests.

PR #31, opened 2017-07-27
Refactor highlighter

ztolley:
The use of the word dynamic wasn't clear what it meant. Wanted to make the helper more generic and reusable. Also ties the helper function to the associated css class.

The use of the strong tag helps to signify to screens readers that it is significant (which is why something is highlighted).

If a user did not provide another researcher then the other researcher section was still shown as it was creating a highlighted section with no text before testing existence.

Also the other researcher property was present but had no text, so needed to be tested for nil and length > 0

Moved the highlight style into the barnardos folder as it's a generic class that is not tied to a single screen but should be reusable throughout a service.

Added a little style to the highlighters just to give them a some polish.

### DIFF
--- a/app/assets/stylesheets/app/session_preview.scss
+++ b/app/assets/stylesheets/app/session_preview.scss
@@ -1,3 +1,0 @@
-.highlight {
-  background: $highlight-background-colour;
-}
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -3,13 +3,12 @@
 @import "barnardos/accessibility";
 @import "barnardos/base";
 @import "barnardos/button";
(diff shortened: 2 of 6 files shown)


ztolley:
I dunno it feels too closely tied to a single requirement. I'd say you are 'highlighting' something to indicate that it is editable

rgarner:
I think I feel the opposite about this – we only have one use case for highlight, that's to indicate things that are editable (those are our only actual application semantics at the moment), and the "why" of the highlight should be obvious from the code. If we try to library-ise this right now we're getting into premature reuse and breaking the [Rule of Three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) (to which I am somewhat attached).

PR #34, opened 2017-07-31
Humanize has_at_least_one_of validator

rgarner:
De-jankify error messages.

Before:

before

After:

screen shot 2017-07-31 at 15 06 27

### DIFF
--- a/spec/validators/has_at_least_one_validator_spec.rb
+++ b/spec/validators/has_at_least_one_validator_spec.rb
@@ -24,7 +24,7 @@ RSpec.describe HasAtLeastOneValidator, type: :validator do
     it 'has an error' do
       expect(ingredient_errors.length).to eql(1)
       expect(ingredient_errors.first).to \
-        eql('ingredients should have at least one of [:bacon, :eggs, :beans]')
+        eql('should have at least one of [:bacon, :eggs, :beans]')
     end
   end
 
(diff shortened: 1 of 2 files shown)


spec/validators/has_at_least_one_validator_spec.rb
```diff
@@ -35,18 +35,18 @@ class Breakfast < Struct.new(:ingredients)
...
-  context 'Our ingredients are not an array' do
+  context 'Our ingredients are not an array (non-human error)' do
     let(:ingredients) { :'A single hash brown' }
 
     it { is_expected.to be_invalid }
     it 'has an error' do
       expect(ingredient_errors.length).to eql(1)
       expect(ingredient_errors.first).to \
-        eql('ingredients should be an enumerable')
+        eql('should be an enumerable')
```
ztolley: Is this what the user would see?

rgarner: Yes, but it shouldn't happen.  It's difficult, because you wouldn't ordinarily expect a validator to throw an exception, but someone needs to know about it. This isn't ideal. I don't currently have a better suggestion.

rgarner: One possibility is to take it out completely. When you do this the error is raised when you attempt to save the record – but it's a bit cryptic, it's a `PG::Error` related to trying to stuff a string into an `ARRAY` field.

rgarner:
Are the before/after screen grabs not loading for you? Admit they're not the best (since to save time I nicked yours from your other branch) but they show the change in text.

ztolley:
ignore comment about before after, i'm an idiot.

PR #36, opened 2017-08-01
Fix issue with textarea error indicator

ztolley:
Previously the style for the error indicator was being selected for all textareas, an additional selector for .has_error needed to be added.


rgarner:
Needs rebase

PR #38, opened 2017-08-01
Normalise content structure

ztolley:
Restructure the use of headings to be more accessible and try and get some consistency on sizing.

Added a few helper classes

Fixed issue with textfield and textarea not allow additional label classes


rgarner:
Can we rebase (and squish the `fixup!` in the process)? Ta

PR #39, opened 2017-08-02
Rubocop (Ruby linter)

rgarner:
# [Add Ruby linter](https://trello.com/c/tFyBnQDN/42-add-a-ruby-linter)

Add rubocop by:

- adding the gem
- generating a `.rubocop-todo.yml`
- inheriting from it in `.rubocop.yml`
- turning on the rubocop rake task in the `Rakefile`
- Introducing one of our biggest offenders (block length) and ignoring it for spec/rake/cucumber
- Fixing hash formatting for Age as a second example

Stopping at this point. We can turn on other checks as we go by removing items one by one from `.rubocop-todo.yml`.


rgarner:
Yeah, s'true, but the plan is to extract this stuff into a general linter
gem a la gov.uk and link to *that* in the devs-manual.

rgarner:
@ztolley ping

PR #40, opened 2017-08-02
Fix line lengths to 100

rgarner:
- This is the github wrap length
- It's not 80, which is pretty difficult to keep to, but neither is it
  120, which is pretty difficult to read

Fire away...

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -12,8 +12,26 @@ AllCops:
     - node_modules/**/*
     - bin/**/*
 
+Metrics/LineLength:
+  Max: 100
+  Exclude:
+    - config/environments/**/*
+    - config/initializers/**/*
+    - db/seeds.rb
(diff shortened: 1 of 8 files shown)


rgarner:
You mean https://github.com/barnardos/consent-form-builder-rails/blob/f0097d4b760354fa41ad1bd20c4c7747d221de56/.rubocop.yml#L18-L23 ?

ztolley:
Them why the changes to form_helpers.rb?

rgarner:
The problem was that the method length was getting out of hand. This PR is trying to constrain line length project-wide and that can't be done without either ignoring method length in helpers or upping the line count. So here we've [upped the line count](https://github.com/barnardos/consent-form-builder-rails/blob/f0097d4b760354fa41ad1bd20c4c7747d221de56/.rubocop.yml#L33)

PR #41, opened 2017-08-02
Preserve research focus linebreaks

rgarner:
[(2) Carry through formatting (i.e. line-breaks) from Form to to Print Format](https://trello.com/c/yA9an8mR/74-2-carry-through-formatting-ie-line-breaks-from-form-to-to-print-format)

As a researcher,
I want my research focus to appear with the minimal formatting I used on the entry page
So that I'm not terribly confused and frustrated

- [x] The formatting I used on the previous Research Focus page has its linebreaks preserved at Session Preview

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,7 @@
 inherit_from: .rubocop_todo.yml
 
 AllCops:
+  TargetRubyVersion: 2.4.1
   Include:
     - Gemfile
     - Rakefile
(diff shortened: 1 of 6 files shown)


rgarner:
Couldn't have been; used the `<<~` heredoc form in this PR, so build would have failed on rubocop task

rgarner:
(Actually, that's rubbish, just tried it without and it fails nothing. It was warning me in the IDE it *would* fail. It wouldn't have)

PR #42, opened 2017-08-02
Add support to lint Ruby with Rubocop on Windows

ztolley:
Rubocop on Windows defaults to check for line endings native to the OS, in the case of Windows this is CRLF, but the project is set to enforce uniform use of LF.
This Rubocop setting makes the linter check for LF no matter what the OS is used for linting.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -35,3 +35,5 @@ Metrics/MethodLength:
 Metrics/ModuleLength:
   Max: 200
 
+Layout/EndOfLine:
+  EnforcedStyle: lf


rgarner:
Can you reword your commit message to use the [50/72 guideline](https://github.com/barnardos/devs-manual/blob/master/styleguides/git.md#example) from the devs-manual?

I'm typically using 80-char terminals side by side and on my machine your commit looks like this:

screen shot 2017-08-03 at 07 32 26

rgarner:
Do you know if it's possible to rename the branch on which this pull is based without closing and re-opening? If we merge this now the 'rubocorp' spelling mistake you kindly corrected everywhere else will show up in the merge commit.

ztolley:
Not that I'm aware of. So either I create a new one or we keep the catchy name that reminds me of Robocop

rgarner:
Holy conflation of Robocop and Omnicorp, Batman

:bat:

rgarner:
Just the commit message reword for 72 width please then, Bob

PR #44, opened 2017-08-03
Set content and form max widths

ztolley:
Previously forms and content stretched the entire width of the 1200px grid, making text hard to read and forms a strange size.

Introduced changes to sizes to make the widths more natural.


rgarner:
Do the headings need any love?

screen shot 2017-08-03 at 15 18 45

ztolley:
hmm

ztolley:
That is a bug to do with positioning of legends, not the width.

We need to raise a separate card for that

PR #45, opened 2017-08-03
Link back to steps from preview

rgarner:
# [(2) Return to Previously Completed Steps from Preview in order to Edit](https://trello.com/c/gXdB0ohR/94-2-return-to-previously-completed-steps-from-preview-in-order-to-edit)

As a researcher
I want to indicate the thing I want to edit in the preview
So that I can go directly to that page

- [x] Editable parts in the session review link back to the wizard step on which you can edit them

## Tech notes

- We've moved things related to wizard steps into a nested class `ResearchSession::Steps`
- There is no class method called `h` before you ask, @sl33p 

## Things to do

- Make it look not godawful (please, Obi-Wan-@ztolley, you are my only hope)

## Bugs encountered on the way (now to be separate PRs)

- [Fix header mismatches in three wizard steps](6b45aa6)
- [Move focus.html.erb label_options together](2fe3ffec5453357f788843235c3a93a371bfdc31)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -37,3 +37,7 @@ Metrics/ModuleLength:
 
 Layout/EndOfLine:
   EnforcedStyle: lf
+
+Style/PredicateName:
+  Exclude:
+    - 'spec/**/*'
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -251,16 +251,6 @@ Style/PercentLiteralDelimiters:
     - 'lib/tasks/cucumber.rake'
     - 'spec/controllers/research_sessions_controller_spec.rb'
 
-# Offense count: 1
-# Configuration parameters: NamePrefix, NamePrefixBlacklist, NameWhitelist.
-# NamePrefix: is_, has_, have_
-# NamePrefixBlacklist: is_, has_, have_
-# NameWhitelist: is_a?
-Style/PredicateName:
-  Exclude:
-    - 'spec/**/*'
-    - 'app/models/research_session.rb'
-
 # Offense count: 80
 # Cop supports --auto-correct.
 # Configuration parameters: EnforcedStyle, SupportedStyles, ConsistentQuotesInMultiline.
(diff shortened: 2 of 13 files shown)


app/views/research_sessions/recording.html.erb
```diff
@@ -1,7 +1,7 @@
 <%= render 'back_link' %>
 <div class="main-content">
   <%= render 'error_summary' %>
-  <h1 class="visually-hidden">Recording Methods</h2>
+  <h1 class="visually-hidden">Recording Methods</h1>
```
ztolley: DOH!

app/views/research_sessions/focus.html.erb
```diff
@@ -1,17 +1,17 @@
 <%= render 'back_link' %>
 <div class="main-content">
   <%= render 'error_summary' %>
-  <h1 class="visually-hidden">Research Focus</h2>
```
ztolley: Somehow this has resulted in 
hint issue

ztolley: Edit _textarea.scss to update the hint class to

```
.textarea__hint {
  color: $secondary-text;
  display: block;
  font-family: $font-family;
  font-size: $base-font-size;
  line-height: 24px;
  margin-top: $gutter / 2;
}
```

May need to do same for other form controls

ztolley:
I'd consider splitting to 2 PR's. One for the card (the link back) and another for the headings because it may be a can or pandoras boxes

app/views/research_sessions/focus.html.erb
rgarner: I *wondered* where that had come from. I suspect it's [this](https://github.com/barnardos/consent-form-builder-rails/pull/45/commits/13eba2025f9028a8e0f4f7e9a1cccd54720a8951)

rgarner:
Ok @ztolley, split off into a separate branch and deleted those two commits. While I have the context, will open a card for these fixes

rgarner:
Headers and hints now in #48

PR #46, opened 2017-08-03
Add Hotjar script for feedback

rgarner:
# [(3) Add Feedback Functionality](https://trello.com/c/fR8jrCcx/77-3-add-feedback-functionality)

**NB: Ad blockers will kill this - disable or whitelist**

Added to `app/javascript/vendor`. It's a single anonymous function that
doesn't need to be referenced anywhere, so it's included with a require
that doesn't save any module reference.

If this is working you'll see a feedback logo at the bottom right of
your browser:

screen shot 2017-08-03 at 15 57 20

### DIFF
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -10,3 +10,5 @@
 const HighlightControl = require('../components/highlightcontrol')
 
 HighlightControl.init()
+
+require('../vendor/hotjar')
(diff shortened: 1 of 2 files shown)


ztolley:
Does it do more than the faces? Does it ask you questions?

sl33p:
@ztolley Yeah, it gives you a feedback form. We can give this a whirl and see if it's the right type of thing for us. There are other (more complicated to setup) options in the Hotjar admin section that wouldn't require code changes to implement.

sl33p:
LGTM. Are you happy @ztolley ?

PR #48, opened 2017-08-07
Fix headers and hints

rgarner:
While making #45, we noticed a couple of problems:

1. A few headers had the wrong closing tags 
2. Some options to a helper were duplicated and causing a "last in wins" scenario. Fixing this necessitated an extra CSS change.

### DIFF
--- a/app/assets/stylesheets/barnardos/_textarea.scss
+++ b/app/assets/stylesheets/barnardos/_textarea.scss
@@ -62,6 +62,9 @@
 .textarea__hint {
   color: $secondary-text;
   display: block;
+  font-family: $font-family;
+  font-size: $base-font-size;
+  line-height: 24px;
   margin-top: $gutter / 2;
 }
 
--- a/app/assets/stylesheets/app/session_preview.scss
+++ b/app/assets/stylesheets/app/session_preview.scss
@@ -1,3 +1,4 @@
-.editable, .editable > p {
+.editable,
+.editable > p {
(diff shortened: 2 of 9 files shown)


app/assets/stylesheets/barnardos/_textarea.scss
```diff
@@ -62,6 +62,9 @@
 .textarea__hint {
   color: $secondary-text;
   display: block;
+  font-family: $font-family;
```
ztolley: will this change also be needed in textfield and other form controls?

rgarner:
Made up-to-date, @ztolley

app/assets/stylesheets/barnardos/_textarea.scss
rgarner: Not sure – I cargo-culted this from your other PR.

rgarner: @ztolley how would I tell? Everything *looks* ok?

ztolley: you'd have to make some fake markup in gallery or something then recreate setting the label to have a title like style

rgarner: Ok. I've made a *lot* of markup in gallery, and put some gash styling in the top so we can see what code applies to which examples. This covers things like hints.

Are there any CSS unit testing frameworks we should be considering? 😛

PR #49, opened 2017-08-07
Warn on exit

ztolley:
Most browsers warn the user that they are leaving a page and have made changes to a form. This small polyfill listens for form changes and adds an event to warn the user if they try to leave the page, unless they submit the form in which case it removes the event handler that does that.

The reason the event is added and taken away is that if we just try to return false or something from that event handler it puts that on the screen. The easiest way to handle things is remove the even handler altogether when you submit a page as it's no longer needed.

### DIFF
--- a/app/javascript/lib/exit_warn.js
+++ b/app/javascript/lib/exit_warn.js
@@ -1,0 +1,28 @@
+let changed = false
+const form = document.querySelector('form')
+
+function beforeUnload (event) {
+  const message = 'You have unsaved changes'
+
+  event = event || window.event
+  // For IE and Firefox
+  if (event) {
+    event.returnValue = message
+  }
+
+  // For Safari
+  return message
+}
+
+if (form) {
+  form.addEventListener('submit', () => {
+    window.removeEventListener('beforeunload', beforeUnload)
+  })
+
+  form.addEventListener('change', () => {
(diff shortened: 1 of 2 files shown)


app/javascript/components/highlightcontrol.js
```diff
@@ -29,9 +29,11 @@ class HighlightControl {
...
 
   static init (container = document) {
-    container.querySelectorAll(`.${wrapperClass}`).forEach((element) => {
-      new HighlightControl(element)
-    })
+    Array
+      .from(container.querySelectorAll(`.${wrapperClass}`))
+      .forEach((element) => {
+        new HighlightControl(element)
+      })
```
rgarner: Is this related to warn on exit at all?

rgarner:
The comment from the PR "The reason the event is added and taken away is that if we just try to return false or something from that event handler it puts that on the screen. The easiest way to handle things is remove the even handler altogether when you submit a page as it's no longer needed." is useful intel we don't want to lose. Can we add it to the commit message?

app/javascript/components/highlightcontrol.js
ztolley: No, it's a bug I found whilst I was running the code on ie

PR #51, opened 2017-08-07
Fix compatibility with IE9-11 js

ztolley:
IE does not allow iteration of nodelist.
Convert nodelist to array to allow iteration.


sl33p:
How did this manifest? Are there steps to reproduce so it can be tested?

PR #52, opened 2017-08-08
Add name, signature & date to consent form

rgarner:
- Remove the old disabled checkbox
- Add some disabled inputs instead
- Make them nice and wide for people to write in

fireshot capture 1 - consent form bui_ - http___127 0 0 1_3001_research-session_preview-consent-form

### DIFF
--- a/app/assets/stylesheets/app/consent_form_preview.scss
+++ b/app/assets/stylesheets/app/consent_form_preview.scss
@@ -1,0 +1,32 @@
+.consent-agreement {
+  display: block;
+  background-color: $grey-medium;
+  padding: $gutter;
+  margin-top: $gutter * 2;
+}
+
+.writing-field {
+  display: inline-block;
+  margin-bottom: 24px;
+  padding-right: $gutter;
+
+  label {
+    display: block;
+    margin-bottom: $gutter * 1.5;
+    margin-top: $gutter;
+  }
+
+  .space-for-writing {
+    border: 0;
+    border-bottom: 1px solid black;
+    width: 100%;
+  }
+
+  &.name {
+    width: 100%;
+  }
+
+  &.signature, &.date {
+    width: 50%;
+  }
+}
--- a/app/views/research_sessions/preview_consent_form.html.erb
+++ b/app/views/research_sessions/preview_consent_form.html.erb
@@ -26,13 +26,20 @@
   </li>
 </ul>
 
-<div class="checkbox-group__choice">
-  <input disabled type="checkbox"
-         id="__print_only_empty_checkbox"
-         class="checkbox-group__input">
-    <label class="checkbox-group__label" for="__print_only_empty_checkbox">
-      I agree to my <%='childʼs' if @research_session.unable_to_consent? %>
-      participation in this research and understand that
-      my childʼs data will be used as explained above.
-    </label>
+<p class="highlight super-text">
+  I agree to my <%= 'childʼs' if @research_session.unable_to_consent? %>
+  participation in this research and understand that
+  my childʼs data will be used as explained above.
+</p>
+
+<%# Note: no whitespace between divs important here %>
+<div class="writing-field name">
+  <label for="name">Name</label>
+  <input disabled name="name" class="space-for-writing">
+</div><div class="writing-field signature">
+  <label for="signature">Signature</label>
+  <input disabled name="signature" class="space-for-writing">
+</div><div class="writing-field date">
+  <label for="date">Date</label>
+  <input disabled name="date" class="space-for-writing">
 </div>
--- a/features/printed_consent_forms.feature
+++ b/features/printed_consent_forms.feature
@@ -13,6 +13,7 @@ Feature:
     And I should see links back to edit things that I provided
     When I click continue
     Then I should see the age-appropriate consent form preview
+    And it should have a place for name, signature and date
     And I should see a way to print it
 
 
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -16,3 +16,4 @@
 @import "barnardos/type";
 
 @import "app/session_preview";
+@import "app/consent_form_preview";
--- a/features/step_definitions/consent_form_preview_steps.rb
+++ b/features/step_definitions/consent_form_preview_steps.rb
@@ -7,3 +7,9 @@ end
 And(/^I should see a way to print it$/) do
   expect(page).to have_tag('a', with: { class: 'print-link' })
 end
+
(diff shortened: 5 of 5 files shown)


app/assets/stylesheets/app/consent_form_preview.scss
```diff
@@ -0,0 +1,40 @@
+.consent {
```
ztolley: I know the rule of 3 thing but its good to create generic styles rather than styles tied to a single use case.

Some of these styles could be renamed and don't necessarily have to happen in a consent class.

app/views/research_sessions/preview_consent_form.html.erb
```diff
@@ -26,13 +26,23 @@
...
-<div class="checkbox-group__choice">
-  <input disabled type="checkbox"
-         id="__print_only_empty_checkbox"
-         class="checkbox-group__input">
-    <label class="checkbox-group__label" for="__print_only_empty_checkbox">
-      I agree to my <%='childʼs' if @research_session.unable_to_consent? %>
-      participation in this research and understand that
-      my childʼs data will be used as explained above.
-    </label>
+<div class="consent">
```
ztolley: Don't need a style for consent.

features/printed_consent_forms.feature
```diff
@@ -13,6 +13,7 @@ Feature:
     And I should see links back to edit things that I provided
     When I click continue
     Then I should see the age-appropriate consent form preview
+    And it should have a place for name, signature and date
```
ztolley: if the css classes are for the test, why not use ID's as there is one one of each on the page.

rgarner: Hmm, if the IDs aren't being used for anything else but testing, perhaps they should come out.

app/views/research_sessions/preview_consent_form.html.erb
rgarner: I'll remove.

app/assets/stylesheets/app/consent_form_preview.scss
rgarner: I'll compromise on rule of two in this case then :wink:?

app/assets/stylesheets/app/consent_form_preview.scss
```diff
@@ -0,0 +1,38 @@
...
+
+  &.name {
+    width: 100%;
+
+    .space-for-writing {
+      width: 97%;
+    }
+  }
+
+  &.signature, &.date {
```
ztolley: I hate having to name bits like this, it's never easy.

On the one hand I'd say find a simple grid helper/thing and use that or come up with a name like .row and col-50

ztolley: And as asked before, why 49.5?

Also why 95 and 97 ?

app/assets/stylesheets/app/consent_form_preview.scss
```diff
@@ -0,0 +1,38 @@
+.consent-agreement {
```
ztolley: Is a generic name or way we could do this?  Could you just use p.highlight?

I only say these things because I come from the school of thought with CSS styles that you create everything as a re-usable component from the start, especially as I want to extract the styles into their own npm package and GEM so I can retrofit them to FGM

rgarner: My standard way of going about this is to extract on the second or third use case such that the context that that second or third use case brings can be applied. It seems to me less possible to do it now because I'll be bringing out a crystal ball as to how it might be applied to a project I've not seen (also I don't feel so qualified to contribute to a house style without that having some origin in @grockett's designs).

That's [this school](http://wiki.c2.com/?PrematureGeneralization) of thought. To me generalisation is something that's done grudgingly, after you've seen the same thing multiple times and need a common pattern for it. Is CSS different? Are there any good articles that can help me understand why?

app/assets/stylesheets/app/consent_form_preview.scss
rgarner: 49.5 because at 50 the date portions breaks onto a separate line. 95 and 97 to keep things lined up using a relative unit (yes, I'm afraid they were arrived at by trial and error). You're hitting the limits of my CSS here; I never knew how to do this properly with transparently addable properties. Little help?

rgarner: > On the one hand I'd say find a simple grid helper/thing and use that or come up with a name like .row and col-50

As far as I've seen from #ui-kit we're working on our own grid as part of elements. I don't want to assign generic names to these because I don't want to:

1. give the idea to devs coming after (by example) that this is in any way *the* grid system. I *want* devs to look at this and say "oh, ok, single use case".
2. design a grid system, even in miniature. My CSS is passably OK, but there are people around here better qualified 😉

rgarner: Ok, weird units eradicated due to realising this is a limitation with [whitespace around `inline-block`](https://css-tricks.com/fighting-the-space-between-inline-block-elements/)

app/assets/stylesheets/app/consent_form_preview.scss
ztolley: I agree that guessing how to generalise things can sometimes not be ideal but where possible, building interface elements as components is the preferred approach. I'm very much of the school of thought that you 'think in components' (There's plenty of posts with this as their title) and there are just as many articles out there that discuss this approach as there are about premature optimisation.

All I'm asking is, when looking at a new thing look for an existing component/pattern that we hve created and if there isn't one then try to think of it in a general way, not specific to one screen or one feature. With the fields that looks good, with the 50/50 thing I'd come up with a name split 'split-panel' or something.

I agree that Gary needs to be involved, and want him to start looking at the PR's so he's aware of the interface work we are doing so he can reuse the patterns we come up with.

In the projects I've worked on, using the component approach meant that after the initial styles were agreed we rarely needed a designer to create screens or designs, and the interfaces defined themselves based on the needs and existing patterns.

rgarner:
That's cool: you're qualified to do so :)

app/assets/stylesheets/app/consent_form_preview.scss
sl33p: I agree that a component based approach is the right way to go. A concern however is that when you (prematurely) define a component that hasn't been through the rigour of a design process and proven as being a truly reusable pattern, you still end up using this component for anything that approximately looks like a fit. 

If there is a way to _revisit the components, and revise their design_ that has a lower overhead than proving that they truly are universal components up front (through a design process over several use cases) then we don't have a problem with defining components in the first use case.

$0.02

PR #54, opened 2017-08-09
Add a link to the current release SHA

rgarner:
- Adds a link to the current `HEAD` commit on Github in a minimal footer
- Dependent on github links
- Dependent on presence of `HEROKU_SLUG_COMMIT` if in production
- Will show 'Release: unavailable' for any environment without
  `HEROKU_SLUG_COMMIT` or a git repo with a `HEAD`

We can do better than this, but this at least gives PM and DM an idea
of what's live. We can get to reasonable release naming and a changelog
next – this ticket is "minimal what's live"

screen shot 2017-08-10 at 12 10 53

### DIFF
--- a/app/assets/stylesheets/barnardos/_page.scss
+++ b/app/assets/stylesheets/barnardos/_page.scss
@@ -27,6 +27,18 @@
   margin-right: $gutter;
 }
 
+#global-header__release {
+  width: auto;
+  max-width: 50em;
+  display: block;
+
+  > label {
+    font-size: 80%;
+    float: right;
+    margin-right: $gutter;
+  }
+}
+
 #global-header-bar {
   background: $primary-colour;
   height: $gutter / 2;
--- a/app/assets/stylesheets/barnardos/_page.scss
+++ b/app/assets/stylesheets/barnardos/_page.scss
@@ -27,18 +27,6 @@
   margin-right: $gutter;
 }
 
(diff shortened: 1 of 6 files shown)


ztolley:
Could this be shown only if being run in development mode or under development environment and not shown for production?

Also, could it not be in the head but maybe a 'stuck' to the bottom right of the screen, or top right?

rgarner:
Any of those things are possible, but the development thing is probably the least useful use case. The real use case here is for @sl33p and Veronique (username unknown) knowing what's on each 'production-like' (ie stage and prod) environment (though me buggering about with Heroku yesterday was a really useful use of a morning). This is a minimal starter for ten.

Another alternative is that we provide a totally separate URL (something like `/version` or `/changelog`) so that the real benefit of knowing what functional things are live is better realised.

In other words, this will come out at some point – is the benefit to be had from fettling it now primarily so that the header looks a bit less technical in show & tells?

rgarner:
(was thinking of chucking something like https://github.com/devtyr/trello-releasenotes at our "Awaiting Deploy" column)

ztolley:
If we use PR branches, the feature is in the url

rgarner:
Didn't know that, and that's helpful, and this PR is about what's in stage/prod when that feature URL isn't there

rgarner:
Lemme try the bottom right...

PR #56, opened 2017-08-10
Add app.json to support Heroku PR deployment

ztolley:
Heroku requries an app.json to tell it how to configure PR created instances, this is that app.json

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,0 +1,15 @@
+{
+  "name":"consent-form-builder-rails",
+  "scripts":{},
+  "env":{
+    "LANG": { "required": true },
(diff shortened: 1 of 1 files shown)


rgarner:
Just looking at the [`app.json` schema](https://devcenter.heroku.com/articles/app-json-schema#example-app-json) – is there *anything* in here that's mandatory? 😆

ztolley:
This was based on an autogenerated one and has all that we need in it

rgarner:
Yeah, sorry, that wasn't a blocking comment, should've just merged at that point and forgot 😄

PR #57, opened 2017-08-10
Make the footer stick to the viewport

rgarner:
Oops. Only resized my browser, so missed this on scroll. Go with
`position: fixed` rather than `absolute`.

screen shot 2017-08-10 at 12 59 36


ztolley:
Did you test on iphone/ipad?

rgarner:
Can confirm working on iPad 3, iOS 9.3.5

PR #65, opened 2017-08-22
Show progress

ztolley:
# Indicate progress through tool

Uses a visual indicator to let the user know where they are in the journey and navigate back and forward as needed.

- Users can always navigate back
- Users can only navigate forward if they have previous filled in a future screen
- Link to the preview page when all sections are complete.

progress

### DIFF
--- a/app/views/research_sessions/_progress.html.erb
+++ b/app/views/research_sessions/_progress.html.erb
@@ -1,23 +1,51 @@
 <div class="progress">
   <ol class="progress-bar">
-    <li class="progress__stage  is-complete">
-      <span class="progress__stage-title">
-        <%= link_to 'Start', start_path %>
-      </span>
-    </li>
-
-    <% wizard_steps.each do |step_key| %>
-      <li class="progress__stage">
+    <% if step %>
+      <li class="progress__stage is-complete">
         <span class="progress__stage-title">
-          <%= link_to t(step_key), question_path(step_key) %>
+          <%= link_to 'Start', start_path %>
+        </span>
+      </li>
+    <% else %>
+      <li class="progress__stage is-active">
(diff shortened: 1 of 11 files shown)


sl33p:
This is looking really sweet. On a design-related note, should the indicators in a case like below (where you've visited and filled in "Recording" and "Focus" already) be solid green?

screenshot 2017-08-22 16 46 26

ztolley:
I like it as it is. The solid shows you where in the journey you are, if you made future ones green because they are filled in it may be confusing and make it unclear where in the journey you jumped to

app/views/research_sessions/_progress.html.erb
```diff
@@ -0,0 +1,51 @@
...
+    <% end %>
+
+
+    <%#
+      If the user has completed the last step, provide
+      a link to the preview section
+    %>
+    <li class="progress__stage">
+      <span class="progress__stage-title">
+        <% if  @research_session and @research_session.reached_step?(wizard_steps[-1]) %>
```
rgarner: `wizard_steps.last` (from [`Array#last`](https://ruby-doc.org/core-2.4.1/Array.html#method-i-last)) would also work here.

config/locales/en.yml
```diff
@@ -30,9 +30,16 @@
...
 
 en:
+
   hello: "Hello world"
+  methodologies: "Methodologies"
+  recording: "Recording"
+  focus: "Focus"
+  researcher: "Researcher"
+  incentive: "Incentive"
+  age: "Age"
```
rgarner: We probably don't need translations for these; [`Symbol#humanize`](http://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-humanize) would do the same.

PR #68, opened 2017-08-25
Update researcher screen

ztolley:
Update the content in the researcher screen.

Add an option to ask if there is another researcher, and if there is reveal the other researcher field.

### DIFF
--- a/db/migrate/20170825104045_add_conditional_researcher.rb
+++ b/db/migrate/20170825104045_add_conditional_researcher.rb
@@ -1,0 +1,5 @@
+class AddConditionalResearcher < ActiveRecord::Migration[5.1]
+  def change
+    add_column :research_sessions, :researcher_other, :boolean
+  end
+end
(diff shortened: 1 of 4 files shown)


db/migrate/20170825104045_add_conditional_researcher.rb
```diff
@@ -0,0 +1,5 @@
+class AddConditionalResearcher < ActiveRecord::Migration[5.1]
+  def change
+    add_column :research_sessions, :researcher_other, :boolean
```
rgarner: If we've added a boolean for another researcher, we also need `presence` validations for the other researcher fields. If we're not collecting those fields any more, we should remove them.

As of 5bfb4095910c0f5378dd03659748e169995d8f71 I can't make the conditional subfield work to add any other researcher details?

rgarner: Oh, ok, is it not supposed to be [working yet](https://trello.com/c/kpHLeXSK/122-add-another-researcher)?

rgarner: Guh, never mind. Forgot to start webpacker server 🆔 iot

rgarner: Ok, deleting irrelevant comments, leaving this relevant one:

If we've added a boolean for another researcher, we also need presence validations for the other researcher fields.

PR #69, opened 2017-08-25
Split the focus field into topic and purpose

ztolley:
Updated design the focus question into 2 separate questions, topic and a question about the purpose of the research.


app/views/research_sessions/topic.html.erb
```diff
@@ -0,0 +1,19 @@
...
+<div class="main-content">
+  <%= render 'error_summary' %>
+  <h1 class="visually-hidden">Research Topic</h1>
+  <%= form_for @research_session, url: wizard_path do %>
+      <%= labelled_text_area_tag(
+            :topic,
+            'What is the research or participation session about?',
+            @research_session.topic,
+            label_options: {
+              hint: 'The topic is the project is …[what is the research about?]',
```
rgarner: Should read "The topic *of*"?

PR #72, opened 2017-08-28
Change age flow

ztolley:


app/views/research_sessions/preview.html.erb
```diff
@@ -1,5 +1,15 @@
 <a href="javascript:window.print()" class="print-link">Print</a>
 <h1 class="title screen-only">Research session preview</h1>
+
+<section class="split-link screen-only">
+  <div class="<%='active' if @research_session.able_to_consent? %>">
+    <a id="over18-link" href="?age=over18">Respondants who can give consent themselves</a>
```
rgarner: typo: Respondents

app/views/research_sessions/preview.html.erb
```diff
@@ -1,5 +1,15 @@
...
+  </div>
+  <div class="<%='active' if @research_session.unable_to_consent? %>">
+    <a id="under18-link" href="?age=under18">Respondants who can't give consent</a>
```
rgarner: typo: Respondents

features/step_definitions/question_steps.rb
```diff
@@ -71,6 +69,8 @@
   fill_in 'Incentive value', with: '10.50'
 
   click_button 'Continue'
+
+  click_link "Respondants who can't give consent"
```
rgarner: typo: Respondents

rgarner:
We should also remove `app/views/age.html.erb`

rgarner:
We've deprecated Age as a concept but left it as a column. It looks like it's just a random query string tied to "able to give consent" now. It looks like we've taken on a little tech debt leaving the concept of "age" kicking around. Should anything be stored in the DB at all here?

PR #78, opened 2017-09-01
Add expenses

ztolley:
Adds the required fields to the DB and a new screen to capture information related to expenses, as described in :

https://docs.google.com/document/d/1hk7eI9rqzeNMxzrkF4UoJUY9IW-gN-ixQCWnZ6mqYTQ/edit

capture


rgarner:
We don't need validations on anything here, do we? If not, LGTM.

(but there are conflicts now 😝 )

PR #80, opened 2017-09-04
Prefer #humanize/inflections over translations

rgarner:
If we use `I18n.translate`, translations have to exist in `en.yml`,
otherwise errors make it into the markup (see below).
`String#humanize` doesn't have this issue
(but still lets us specify our own translations).

Before, in _progress output (because we didn't have `en.time_equipment`):
```HTML
<a href="/questions/time_equipment">
  <span class="translation_missing"
        title="translation missing: en.time_equipment">
        Time Equipment
  </span>
</a>
```

After, including a `#human` call in `initializers/inflections.rb`:

```HTML
<a href="/questions/time_equipment">Time / equipment</a>
```

Because the rest of the progress titles are conventional, (for example,
`:methodologies.to_s.humanize => 'Methodologies'`) we don't need
special cases for anything else and can remove the obligation to
translate in `en.yml`.

### DIFF
--- a/app/views/research_sessions/_progress.html.erb
+++ b/app/views/research_sessions/_progress.html.erb
@@ -25,9 +25,9 @@
           <%= 'is-active' if this_step == step %>">
         <span class="progress__stage-title">
           <% if this_step != step and @research_session and @research_session.reached_step?(this_step) %>
-            <%= link_to t(this_step), question_path(this_step) %>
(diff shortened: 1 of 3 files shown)


rgarner:
We don't need validations on anything here, do we? If not, LGTM.

(but there are conflicts now 😝 )

PR #81, opened 2017-09-05
Add step links to preview

ztolley:
Added parameter to preogress indicator to allow indication that we are on the preview stage.

Updated progress indicator to correctly show if the preview screen is the current page.

### DIFF
--- a/app/views/research_sessions/preview.html.erb
+++ b/app/views/research_sessions/preview.html.erb
@@ -1,3 +1,5 @@
+<%= render "research_sessions/progress", preview: true %>
+
 <a href="javascript:window.print()" class="print-link">Print</a>
 <h1 class="title screen-only">Research session preview</h1>
 
--- a/app/presenters/research_session_presenter.rb
+++ b/app/presenters/research_session_presenter.rb
@@ -2,7 +2,7 @@ class ResearchSessionPresenter < Struct.new(:research_session)
   delegate :age, :topic, :purpose, :researcher_name, :researcher_other_name,
            :researcher_email, :researcher_phone, :unable_to_consent?,
            :able_to_consent?, :other_methodology, :other_recording_method,
-           to: :research_session
+           :reached_step?, to: :research_session
 
   def methodology_list
     paras = research_session.methodologies.map do |methodology|
--- a/app/views/research_sessions/_progress.html.erb
+++ b/app/views/research_sessions/_progress.html.erb
@@ -1,16 +1,14 @@
-<div class="progress">
+<% preview ||= false %>
+
+<div class="progress screen-only">
   <ol class="progress-bar">
-    <% if step %>
-      <li class="progress__stage is-complete">
-        <span class="progress__stage-title">
-          <%= link_to 'Start', start_path %>
-        </span>
-      </li>
-    <% else %>
-      <li class="progress__stage is-active">
-        <span class="progress__stage-title">Start</span>
-      </li>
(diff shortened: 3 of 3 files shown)


app/controllers/research_sessions_controller.rb
```diff
@@ -30,7 +30,7 @@ def preview
       @research_session.assign_attributes(params.permit(:age))
     end
 
-    @research_session = ResearchSessionPresenter.new(@research_session)
+    @preview = ResearchSessionPresenter.new(@research_session)
```
rgarner: Rails-y suggestion: leave this as `@research_session` (don't add a separate `@preview` field) and don't check for the existence of `@preview`. Parameterise the _progress partial instead. At https://github.com/barnardos/consent-form-builder-rails/pull/81/files#diff-6e7794c2e7657c77f69b1250c49e3774R1 , say

```ruby
<%= render "research_sessions/progress", locals: { preview: true } %>
```

and in the _progress partial itself, default `preview` to `false` by saying

`<% preview ||= false %>`

Then elsewhere in _progress you'll have a local variable preview which you can use as a variable local to that partial:

```
 <li class="progress__stage <%= 'is-active' if preview %>">
```

Lastly, add `:reached_step?` to the delegate list in the presenter.

ztolley: it was left as research session before and I tried adding reached_step? to it but it was not having it. It's maybe due to steps and session both having a reached_step method (think that may need work).

I'll revert it to what you are suggesting to demonstrate it failing when I come in.

ztolley: actually doing it that way means I don't need to extend the delegate, but it did feel a bit wrong to put it there.

ztolley: My original reason for not passing a param and looking at session state was it feels like it's mixing having logic built into the fragment that looks at state and having parameters sent into it, ideally it would be all passed as params

app/views/research_sessions/preview.html.erb
```diff
@@ -1,3 +1,5 @@
+<%= render "research_sessions/progress", preview: true %>
```
rgarner: Oh, nice, you don't need `locals` any more? Did not know that.

rgarner:
On Incentive we have lots of green filled circles:

screen shot 2017-09-07 at 14 00 13

But when we hit preview they mostly go hollow:

screen shot 2017-09-07 at 14 01 10

rgarner:
Just noticed that the progress bar shows up in the print stylesheet:

screen shot 2017-09-07 at 15 17 32

rgarner:
Fix yer `or` (or leave it if you prefer, Rubo hasn't complained) and squash yer `fixup!` and we're there?

ztolley:
The visibility in the print is an issue, but I was going to fix that in a follow up as the other PR will be changing styles.

ztolley:
As for 'or', I had use that because we use 'and' and didn't want to mix the two, plus I'm used to nunjucks which only allows and/or

ztolley:
Changed the syntax, but will address the print issue in a follow up.

rgarner:
Ok, just squash the fixup then please.

rgarner:
Just noticed the commit message in https://github.com/barnardos/consent-form-builder-rails/pull/81/commits/d9261e158a83b49bbe0f5a21e78604bfd7ea4911 still references the thing we were doing before (but aren't now)

PR #89, opened 2017-09-12
Improve page spacing consistency

ztolley:
screenshot from 2017-09-12 11-54-32

* Adjusted the progress bar to result in a consistent height.
*  Swited page headings to consistent structure for consistent heading sizes and spacing
* Adjusted radio headings to fix spacing oddity.

### DIFF
--- a/app/assets/stylesheets/barnardos/_checkbox.scss
+++ b/app/assets/stylesheets/barnardos/_checkbox.scss
@@ -17,7 +17,7 @@ $outer-ring: $inner-ring + ($ring-size * 3);
 }
 
 .checkbox-group__legend {
-  font-size: 23px;
+  font-size: $base-font-size;
   font-weight: normal;
   line-height: 24px;
   margin: ($gutter / 2) 0;
--- a/app/assets/stylesheets/barnardos/_fieldset.scss
+++ b/app/assets/stylesheets/barnardos/_fieldset.scss
@@ -6,7 +6,7 @@
 }
 
 .fieldset__legend {
-  font-size: 20px;
+  font-size: 18px;
   font-weight: normal;
   line-height: 32px;
   margin: 12px 0;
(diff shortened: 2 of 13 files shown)


rgarner:
Just sped through the screens and things definitely *feel* more together, nowhere more so than on the preview page.

features/step_definitions/question_steps.rb
```diff
@@ -2,15 +2,12 @@
...
-  within '[name="first-researcher"]' do
-    fill_in 'Full name', with: 'Rachel Researcher'
-    fill_in 'Telephone number', with: '012345678'
-    fill_in 'Email', with: 'rachel@researcher.com'
-  end
+  fill_in 'researcher_name', with: 'Rachel Researcher'
+  fill_in 'researcher_phone', with: '012345678'
+  fill_in 'researcher_email', with: 'rachel@researcher.com'
+
+  fill_in 'researcher_other_name', with: 'Steve Secondresearcher'
```
rgarner: Can we stick to filling in fields by labels? There are two good reasons for doing this:

1. It shows that our labels are consistently tied to their fields (this was very useful indeed in `move-to-form-with` – I could see what linkages I'd broken that weren't covered in `spec`)
2. There's a big conflict coming down the pipe from that very branch :wink:

I can't see the reason for changing this unless it's "well I changed a few labels and I don't want to change this test again", which is on the face of it a good reason but for 1. above.

features/step_definitions/question_steps.rb
```diff
@@ -83,15 +80,12 @@
...
```
rgarner: Again, labels are preferable.

ztolley:
So that's a thumb up and approve?

features/step_definitions/question_steps.rb
ztolley: We can't use the label because there are two fields on the screen with 'first name'

rgarner: There are two "Full name" fields, yes, and we can because we have (or at least had) the `within` blocks

rgarner: (you're quite right to say we can't go by label alone; we previously had to reference the fieldset by name in a `within`, is what I'm trying to say)

PR #95, opened 2017-09-18
Convert to factory_girl

rgarner:
There was a lot of [duplication](https://github.com/barnardos/consent-form-builder-rails/blob/master/spec/models/research_session_spec.rb#L279-L289) in `research_session_spec.rb` with respect to setting up models. Suddenly we need to do the same kind of thing again in order to set up tests to [redirect to partially-completed forms](https://trello.com/c/Su0Kmc51/150-make-urls-returnable#comment-59bfa06495ff1052eadba0d3).

Rather than pollute `research_controller_spec.rb` with similar duplication, make a factory for each `ResearchSession::Steps` step.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,6 +29,7 @@ group :test do
   gem 'cucumber-rails', require: false
   gem 'database_cleaner'
   gem 'launchy'
+  gem 'factory_girl_rails'
 end
 
 group :development do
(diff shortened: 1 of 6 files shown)


rgarner:
Any comment, @ztolley? Hopefully this should help with you testing `Researcher`s

PR #96, opened 2017-09-19
Tidy up the inline error styles

joelcarr:
# [Tidy up the inline error styles](https://trello.com/c/qmwS3Q2b/164-tidy-up-the-inline-error-styles)

Tidying up the error state styles for textfields and textareas that are present in the screenshot featured in https://github.com/barnardos/consent-form-builder-rails/pull/93. These updates bring the styles more inline with the styleguide.

## Screenshot:

image

### DIFF
--- a/app/assets/stylesheets/barnardos/_textarea.scss
+++ b/app/assets/stylesheets/barnardos/_textarea.scss
@@ -1,8 +1,6 @@
 @import "barnardos/colours";
 @import "barnardos/vars";
 
-$error-icon-size: 22px;
-
 .textarea {
   margin: $gutter 0 $gutter (-$gutter / 2);
   max-width: $max-form-control-width;
@@ -27,7 +25,7 @@ $error-icon-size: 22px;
   margin-top: $gutter / 2;
   position: relative;
 
-  .has-error > & {
+  .has-error & {
     position: relative;
     padding-right: ($error-icon-size + $gutter + $gutter / 2);
 
--- a/app/assets/stylesheets/barnardos/_textarea.scss
+++ b/app/assets/stylesheets/barnardos/_textarea.scss
@@ -1,10 +1,12 @@
 @import "barnardos/colours";
 @import "barnardos/vars";
 
+$error-icon-size: 22px;
+
 .textarea {
   margin: $gutter 0 $gutter (-$gutter / 2);
   max-width: $max-form-control-width;
-  padding: $gutter / 2;
+  padding: ($gutter / 2) ($gutter / 2) $gutter;
   position: relative;
 }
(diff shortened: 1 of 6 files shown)


ztolley:
Shouldn't we add a ! next to the label for radio and checkbox?

app/assets/stylesheets/barnardos/_textarea.scss
```diff
@@ -23,21 +23,28 @@
   color: $black;
   font-weight: normal;
   margin-top: $gutter / 2;
+  position: relative;
+
+  .has-error > & {
```
rgarner: Doesn't the `&` mean "parent selector"? I'm a little curious. What does this mean in this context? Not easily googleable 😄

joelcarr: For SASS, the `&` is used when nested to define where the parent class is placed. This will output in the CSS:

```css
.has-error > .textarea__label {
  ...
}
```

You can read some tips about it here: https://css-tricks.com/the-sass-ampersand/
This is also a decent tool for testing out fun SASS things: https://www.sassmeister.com/

joelcarr:
@ztolley we should add the (!) icon to both the radio and checkbox form too.

app/assets/stylesheets/barnardos/_textarea.scss
rgarner: Did *not* know how powerful this was, thanks. I've only been using it to refer to the parent in inner blocks, had no idea you could use it like [this](http://www.joeloliveira.com/2011/06/28/the-ampersand-a-killer-sass-feature/)

joelcarr: It is brilliant for BEM structure as you can do:

```scss
.block {
  display: block;

  &__element {
    color: blue;

    &--modifier {
      background: orange;
    }
  }
}
```

Which will output:

```css
.block {
  display: block;
}

.block__element {
  color: blue;
}

.block__element--modifier {
  background: orange;
}
```

ztolley: Some devs discourage use of that notation with BEM because they would rather you don't nest (as you don't need to) and having the full version makes it easier to find but I'm all for what is practicle.

joelcarr:
Added the error icon for the checkboxes / radios legends too. Here are some screenshots:

image

image

rgarner:
@joelcarr thanks, [bug](https://trello.com/c/Yz48ihXf/177-bug-if-you-dont-select-from-a-list-of-checkboxes-when-thats-the-only-input-you-get-this-error) opened.

PR #99, opened 2017-09-20
Allow tabbing in checkboxes

rgarner:
# [Tabbing through checkboxes isn't working](https://trello.com/c/AtwZN5N8/136-tabbing-through-checkboxes-isnt-working)

screen shot 2017-09-20 at 12 35 46

To restore tab functionality, swap `display: none` in the
`checkbox-group__input` for an `absolute` positioning within its div and an
opacity of 0 such that it responds once again to browser events.

Then style `:focus` the same as `:hover` for `.checkbox-group__input`, and
combine the styles for a selected focussed check box.

### DIFF
--- a/app/assets/stylesheets/barnardos/_checkbox.scss
+++ b/app/assets/stylesheets/barnardos/_checkbox.scss
@@ -56,7 +56,8 @@ $outer-ring: $inner-ring + ($ring-size * 3);
 }
 
 .checkbox-group__input {
-  display: none;
+  opacity: 0;
+  position: absolute;
 }
 
 .checkbox-group__input + .checkbox-group__label {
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/barnardos/_checkbox.scss
```diff
@@ -38,7 +38,8 @@ $outer-ring: $inner-ring + ($ring-size * 3);
 }
 
 .checkbox-group__input {
-  display: none;
+  opacity: 0;
+  position: absolute;
```
joelcarr: Are we being cautious for older browsers yet?

rgarner: It's a good question. I was rather hoping either you or @ztolley could advise me on the contemporaneous viability of this ploy.

rgarner: (alternative would seem to be a left: -999px or somesuch)

ztolley: We don't know the browser, but we are aiming this at Barnardos stuff so worst case, find out what the Barnardos desktops run.

Where did the fix come from?

rgarner: Me.

rgarner: [Compatibility](https://developer.mozilla.org/en-US/docs/Web/CSS/opacity#Browser_compatibility) doesn't seem bad

rgarner: 'k, `left: -9999px` is also perfectly fine, but you lose one small dev advantage – rolling over the 'hidden' `input` in the Inspect mode no longer indicates visually which field we're talking about.

joelcarr: As mentioned in the standup, I think we can leave this as is for now, and come back to it if we need to support IE8 as well. That will probably be a separate bigger ticket anyway.

PR #100, opened 2017-09-20
Fix links in preview returning to steps

rgarner:
# [BUG: Preview page links should return to the appropriate page of the Consent Form Builder for editing](https://trello.com/c/p6HAGJRa/176-bug-preview-page-links-should-return-to-the-appropriate-page-of-the-consent-form-builder-for-editing)

The rails `_path` method here *can* accept a model, but the trouble is
it's really getting a presenter that doesn't behave in quite the same
way. Since only the `ResearchSession#id` is used and the presenter already
has that, pass that through instead

(and update the test double to respond to `:id`)


rgarner:
Believed this had introduced a regression, but was unrelated.

PR #101, opened 2017-09-20
Fix header width on mobile size viewports

joelcarr:
# [Header section causing horizontal scroll](https://trello.com/c/IukVaf3c/100-2-header-section-causing-horizontal-scroll)

On smaller viewports, the header was causing the page to overflow horizontally causing a scroll to black space. Adding the gutter space with padding allows the space to be contained within the width of the element.

### DIFF
--- a/app/assets/stylesheets/barnardos/_page.scss
+++ b/app/assets/stylesheets/barnardos/_page.scss
@@ -20,7 +20,7 @@
 }
 
 #global-header__logo {
-  margin-left: $gutter;
+  padding-left: $gutter;
 }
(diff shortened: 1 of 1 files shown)


rgarner:
LGTM, local Chrome in restricted viewport mode and Nexus 5x

rgarner:
(apologies for the self-request, I couldn't work out where the review tools were, again. That was me thrashing)

rgarner:
This will need a rebase due to our current settings. PRs are currently bottlenecked on this.

PR #104, opened 2017-09-22
Ability to print consent form and info sheets separately

joelcarr:
# [Ability to print consent form and info sheets separately](https://trello.com/c/EhJjHEjS/158-ability-to-print-consent-form-and-info-sheets-separately)

Taking advantage of `@media print {...}` styles, this PR implements a dropdown where the user can choose which area of the completed form preview page they wish to print.

## Screenshot:
image

image

image


ztolley:
How about different buttons instead of a select?

rgarner:
I like the select tbh. Gives us the opportunity to optimise for the common case (assuming we have one!)

rgarner:
This is nice. I have a few questions (not about the code):

- Is the Print button conforming to a Zeplin style somewhere? The hover is nice but it sort of appears disabled
- Should the print button be right-aligned? It seems to have moved
- Is the trailing blank page on all of the three options avoidable?

joelcarr:
@ztolley I had originally done buttons but it felt overwhelming:

image

@rgarner:
- The button is not conforming to a zeplin style. I merged the original print link with a secondary button. Couldn't see a button with text and a icon. 
- The print button was right of the heading however with the select box now it doesn't fit.
- I haven't seen a trailing blank page. Could you provide some steps to reproduce this? Which browser are you using?

rgarner:
@joelcarr I get a trailing blank page in Chrome for every page. Perhaps it's content-dependent. I race through the pages putting in almost the absolute minimum, so one-character fields, a couple of methodologies/recording methods, straight Continue through the options. I'll see if I can make it happen consistently.

rgarner:
Yeah, it's content-dependent. Here's what you can do to trigger it:

1. A, <blank>, @, Continue
2. T, Continue
3. P, Continue
4. Select top 2 opts, Continue
5. Select top 2 opts, Continue
6. Just the team, 1wk, A, Continue
7, 8, 9. Continue
10. Print (anything). Hopefully observe trailing page!

I realise this is likely to be an absolute bugger to fix, so my instinct is to say "just don't worry about it rn"

rgarner:
@joelcarr with respect to the button thing, one or two ideas spring to mind:

1. Keep the text/icon black on hover
2. Mimic the behaviour of our "action" buttons – go dark green and make the text and printer white?

The presence or absence of either of these would not stop me from merging this.

joelcarr:
@rgarner I'll give these a go for the morning. If it is proving too much then we can look into making tickets for later.

joelcarr:
@rgarner I could reproduce the extra page break with your steps. I tried a few different things but couldn't find a way to stop it from happening. Would be good to see if it happens with real data.

The printer icon is being used as a background image making it not easy to change the style of it without changing the implementation. I would suggest going to use [inline SVGs](https://css-tricks.com/using-svg/#article-header-id-7) but that would be a bigger chunk of work worth its own ticket.

joelcarr:
Updated with new styles approved by @sl33p.

image

image

PR #105, opened 2017-09-25
Try Percy

ztolley:
Percy.io adds tha ability to carry out visual regression tests.
This change adds the percy GEM and saves snapshots when going
through the test flow.

Tests run as normal in Circle CI with 2 environment variables set
to tell it which project in Percy.io to save the screenshots to


rgarner:
screen shot 2017-09-25 at 12 39 49

ztolley:
I need @sl33p to help set up an official Barnardos account to use for this, currently using temporary private one.

joelcarr:
Awaiting setting up an account with digital@barnardos.org.uk

features/step_definitions/question_steps.rb
```diff
@@ -59,20 +68,24 @@
...
   fill_in 'If you allow food expenses, what is the maximum allowed?', with: '20.00'
+  Percy::Capybara.snapshot(page, name: 'Expenses')
   click_button 'Continue'
 
   choose 'Yes'
 
   choose 'Cash'
   fill_in 'Incentive value', with: '10.50'
-
+  Percy::Capybara.snapshot(page, name: 'Incentive')
```
rgarner: Would we be ok to use Symbols for this, just so that it's obvious that they mean the same thing as the symbols in `ResearchSession::Steps::PARAMS`?

ztolley: ??? Not sure. Isn't this ok? how would I get the symbols into it... should you Even use symbols. What do you think @joel?

rgarner: I mean just use `name: :incentive` instead of `name: 'Incentive'` 😄

rgarner: (then it's easier mentally to make the connection with the [steps](https://github.com/barnardos/consent-form-builder-rails/blob/24f470d2336b2c969760a429db9233e209d85fa4/app/models/research_session/steps.rb#L8-L18), that's all)

rgarner:
Ok, so I'm running this locally and I've just run my features. Have I sent stuff to Percy, or is it just for CI?

ztolley:
No, it only sends them when running in Circle as it relies on environment variables to specify the Percy login

rgarner:
So can I branch this branch and break all the CSS and then push to see my terrible, terrible work?

I am going to try this now 🚀

ztolley:
Well you can already see difference because EVERY COMMENT HAS THAT SHA

rgarner:
Yeah, I want to see other differences though. It's ok if we see other differences, isn't it? No-one said we were going steady with just the SHA 😄

PR #107, opened 2017-09-26
Preview everything

rgarner:
# [Validate that all relevant content from the form is reflected in the preview copy](https://trello.com/c/lxrTEvdo/157-5-validate-that-all-relevant-content-from-the-form-is-reflected-in-the-preview-copy)

screen shot 2017-09-26 at 12 27 26

## The purpose of this PR ...

... is to provide an area where we can argue about copy / the behaviour of the preview page and verify that copy for the happy path. To that end, it's added some checks testing the eventual copy, and it's probably an improvement over what currently happens in production, but it's a work in progress. **It shouldn't be merged** until we've had a chance to talk about things.

The commits happen pretty much in order of screen flow. The first commit adds some null checks on the preview content (because we weren't checking much on that page).

The following commits fill those checks in.

### DIFF
--- a/app/presenters/research_session_presenter.rb
+++ b/app/presenters/research_session_presenter.rb
@@ -1,4 +1,6 @@
 class ResearchSessionPresenter < Struct.new(:research_session)
+  include ActionView::Helpers::NumberHelper
+
   ResearchSession.attribute_names.each do |attribute|
     delegate attribute, to: :research_session
   end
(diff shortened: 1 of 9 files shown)


app/presenters/research_session_presenter.rb
```diff
@@ -26,4 +30,40 @@ def recording_methods_list
...
+
+  def expenses_sentence
+    return nil unless any_expenses?
+    fragments = [
+      :travel_expenses,
+      :food_expenses,
+      :other_expenses
+    ].map do |expense_attr|
+      value = send("#{expense_attr}_limit")
+      if value&.nonzero?
```
ztolley: Always feels like stuff like this should happen in the view layer.

app/views/research_sessions/preview.html.erb
```diff
@@ -193,6 +212,7 @@
     my childʼs data will be used as explained above.
   </p>
 
+
```
joelcarr: To be super fussy, this line break isn't needed.

app/presenters/research_session_presenter.rb
joelcarr: Do you mean to do the `map` in the view layer too? I am impartial as I can't tell what it would look like in the view.

ztolley: Just feels like something that belongs in a view template

rgarner: As a `Presenter`, this is view layer. Outside of the `html.erb` files, you can either do pure Ruby view logic in `Helper` modules (as we have with [`edit_link_for`](https://github.com/barnardos/consent-form-builder-rails/blob/7819c06e7834ccd853f2f81b3691f7046a04c51a/app/helpers/research_sessions_helper.rb#L2), or – as in this case where the view logic relates to a specific instance of model class `ResearchSession` – by wrapping that `ResearchSession` in a `*Presenter` decorator.

I've been trying to find the best article about the pattern but they're all pretty old now...

app/views/research_sessions/preview.html.erb
rgarner: I'm surprised rubocop hasn't picked it up. I'll get rid.

PR #115, opened 2017-09-28
Add reference to child for respondents who cant give consent

joelcarr:
# [Dynamic Preview Copy - Can or Can't give consent](https://trello.com/c/uexOizmX/146-dynamic-preview-copy-can-or-cant-give-consent)

The info sheet only refers to the reader as the participant/respondent.
Adding in conditional copy that makes it clear who the participant is
and what the research is asking of them depending if they can give
consent.

Will run through with Jess to ensure it makes sense. Can you see any other bits that might be missing?

### DIFF
--- a/app/views/research_sessions/preview.html.erb
+++ b/app/views/research_sessions/preview.html.erb
@@ -66,23 +66,22 @@
       <%= @research_session.methodology_list %>
     <% end %>
     <p>
-      With your permission we will record <%= @research_session.unable_to_consent? ? 'your childʼs/the childʼs in your care' : 'the' %> session using
+      With your permission we will record <%= @research_session.unable_to_consent? ? 'your childʼs/the child in your care‘s' : 'the' %> session using
       <%= edit_link_for(:recording_methods) { @research_session.recording_methods_list } %>.
       Recording the session helps the researcher and other research team members
       trying to improve the service, as it allows them to review the most useful
-      parts of the session after it hasfinished.
+      parts of the session after it has finished.
     </p>
   </section>
   <section>
     <h3 class="subtitle-small" id="doi">Do I have to take part?</h3>
     <p>
       Taking part is entirely voluntary - it is up to you to decide whether or not <%= you_or_your_child %>
-      to take part. If you decide  <%= 'your child/the child in your care' if @research_session.unable_to_consent? %> to take part <%= you_or_they %> do not have to answer questions
-      <%= you_or_they %> do not want to answer. <%= you_or_they.capitalize %> can also change <%= @research_session.able_to_consent? ? 'your' : 'their' %> mind about taking part
+      should take part. If you decide <%= you_or_your_child %> should take part <%= you_or_they %> do not have to answer questions <%= you_or_they %> do not want to answer. <%= you_or_they.capitalize %> can also change <%= @research_session.able_to_consent? ? 'your' : 'their' %> mind about taking part
       at any time and withdraw without giving a reason.
     </p>
     <p>
-      If you agree <%= 'for your childʼs/the childʼs in your care' if @research_session.unable_to_consent? %> to take part, you will be given a copy of this information
+      If you agree <%= 'for your child/the child in your care' if @research_session.unable_to_consent? %> to take part, you will be given a copy of this information
       sheet and be asked to sign a consent form.
     </p>
   </section>
--- a/app/views/research_sessions/preview.html.erb
+++ b/app/views/research_sessions/preview.html.erb
@@ -24,7 +24,7 @@
   <section>
     <h2 class="heading-medium">About this research project</h2>
     <p class="super-text">This information sheet will tell you more about the
-    research so that you can decide whether or not you would like to take part.</p>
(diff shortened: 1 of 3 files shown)


app/views/research_sessions/preview.html.erb
```diff
@@ -66,7 +66,7 @@
       <%= @research_session.methodology_list %>
     <% end %>
     <p>
-      With your permission we will record the session using
+      With your permission we will record the <%= 'your child\'s/the child\'s in your care' if @research_session.unable_to_consent? %> session using
```
rgarner: Need to lose the "the" in the case of `unable_to_consent?` as this'll result in "the your child's".

You can also switch to double quotes here to avoid escaping the apostrophes. Or better still, go to a smart quote: "child‘s" (ALT-])

app/views/research_sessions/preview.html.erb
```diff
@@ -76,13 +76,13 @@
   <section>
     <h3 class="subtitle-small" id="doi">Do I have to take part?</h3>
     <p>
-      Taking part is entirely voluntary - it is up to you to decide whether or not
-      to take part. If you decide to take part you do not have to answer questions
-      you do not want to answer. You can also change your mind about taking part
+      Taking part is entirely voluntary - it is up to you to decide whether or not <%= @research_session.able_to_consent? ? 'you' : 'your child/the child in your care' %>
```
rgarner: I'd be inclined to DRY these up by putting a method in `ResearchSessionsHelper`. Rails is opinionated on getting this stuff out of the view, so you could either add it to the `ResearchSessionsPresenter`, in which case it might look like `<%= @research_session.you_or_your_child %>` but it's terser and more readable (despite the lack of OO purity!) to just add it to `ResearchSessionsHelper` and reference `@research_session` therein:

`<%= you_or_your_child %>`

and in `ResearchSessionsHelper`:

```
  def you_or_your_child
    @research_session.able_to_consent? ? 'you' : 'your child/the child in your care'
  end
```

app/views/research_sessions/preview.html.erb
```diff
@@ -76,13 +76,13 @@
...
+      to take part. If you decide <%= @research_session.able_to_consent? ? 'you' : 'your child/the child in your care' %> to take part <%= @research_session.able_to_consent? ? 'you' : 'they' %> do not have to answer questions
```
rgarner: This will result in some weird English, e.g. "If you decide you to take part" or "If you decide your child/the child in your care to take part"

app/views/research_sessions/preview.html.erb
```diff
@@ -76,13 +76,13 @@
...
+      <%= @research_session.able_to_consent? ? 'you' : 'they' %> do not want to answer. <%= @research_session.able_to_consent? ? 'You' : 'They' %> can also change <%= @research_session.able_to_consent? ? 'your' : 'their' %> mind about taking part
```
rgarner: I'd be inclined to move these to a `ResearchSessionsHelper#you_or_they` and for the start of the sentences say `<% you_or_they.capitalize %>`

app/views/research_sessions/preview.html.erb
joelcarr: Yeah was taking me a while to figure this one out.

app/views/research_sessions/preview.html.erb
```diff
@@ -114,7 +114,7 @@
   </section>
   <% if @research_session.start_datetime.present? or @research_session.duration.present? %>
     <section>
-      <h3 class="subtitle-small" id="private">When is the session and what should I bring?</h3>
+      <h3 class="subtitle-small" id="private">When is the session and what should <%= @research_session.able_to_consent? ? 'I' : 'My child/the child in my care' %> bring?</h3>
```
rgarner: Think the "M" in "my" here should be lowercase.

app/views/research_sessions/preview.html.erb
```diff
@@ -76,13 +76,13 @@
...
-      to take part. If you decide to take part you do not have to answer questions
-      you do not want to answer. You can also change your mind about taking part
+      Taking part is entirely voluntary - it is up to you to decide whether or not <%= @research_session.able_to_consent? ? 'you' : 'your child/the child in your care' %>
+      to take part. If you decide <%= @research_session.able_to_consent? ? 'you' : 'your child/the child in your care' %> to take part <%= @research_session.able_to_consent? ? 'you' : 'they' %> do not have to answer questions
+      <%= @research_session.able_to_consent? ? 'you' : 'they' %> do not want to answer. <%= @research_session.able_to_consent? ? 'You' : 'They' %> can also change <%= @research_session.able_to_consent? ? 'your' : 'their' %> mind about taking part
       at any time and withdraw without giving a reason.
     </p>
     <p>
-      If you agree to take part, you will be given a copy of this information
+      If you agree <%= 'for your child\'s/the child\'s in your care' if @research_session.unable_to_consent? %> to take part, you will be given a copy of this information
```
rgarner: We need to remove the `'s`s here, otherwise we get "If you agree for your child's/the child's in your care to take part"

PR #116, opened 2017-09-29
Add another

ztolley:
# Add Another Researcher

https://consent-form-builder-st-pr-116.herokuapp.com/research-sessions/1/questions/researcher

## Requirements
The premise is simple.  

> As a user creating a research session
> And I am on the 'Researcher' screen
> Then I can define one or more researchers associated with this research session

## Solution
Instead of having the researcher and researcher_ other in the research session table, we create a new table 'researchers' and create a one to many relationship.

When the user goes to the researcher form it will show the single researcher fields it did previously, but now has an 'add another researcher' button to allow more than one to be created.

When the user reaches the preview page it will loop through the researchers.

The add another feature is provided by a javascript component that clones the page element and resets its values. The form data is handled using nested Rails forms.

add-another

## Tasks
- [x] Update model to create session -> researchers
- [x] Update researcher screen to show a single researcher
- [x] Add a new 'AddAnother' js control to allow a fragment of the page to be tagged and an add button created. When the add button is pressed the fragment is cloned and values reset.
- [x] Add some tests for the new control
- [ ] Update validation to require at least one researcher
- [ ] Check form validation
- [ ] Check form data population
- [ ] Update the preview page to support multiple researchers and drop researcher_other
- [ ] Update RSpec tests (part done)
- [ ] Update cucumber tests
- [ ] Add a field to researcher to ask for researcher role

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -10,6 +10,7 @@
 @import "barnardos/datefield";
 @import "barnardos/errorsummary";
 @import "barnardos/fieldset";
+@import "barnardos/fieldset-list";
 @import "barnardos/page";
 @import "barnardos/print-section";
 @import "barnardos/progress";
--- a/app/assets/stylesheets/barnardos/_button.scss
+++ b/app/assets/stylesheets/barnardos/_button.scss
@@ -65,6 +65,32 @@
   }
 }
 
+.button--danger {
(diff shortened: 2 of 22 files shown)


rgarner:
Shelving for now but leaving the branch here as it'll probably come back – prioritised as low after conversation with policy people.

PR #117, opened 2017-10-04
Changelog and rudimentary serving from /changelog

rgarner:
Add a CHANGELOG.md and serve if from `/changelog`

Comes with no tests at present, given that this was for a little spike.
Add a CHANGELOG.md using the suggested changelog standard for the
vandamme gem (used by Gemnasium). It's not particularly mature, but
it's less important that we agree on its content; more that we agree
on something that communicates our changes.

screen shot 2017-10-04 at 12 03 54


sl33p:
🍛

PR #122, opened 2017-10-05
add styles for select fields

joelcarr:
Created styles for the select fields from the forms designs. They
haven't been applied to the date/time select field due to its'
dependency with the rails form builder helper.

image

### DIFF
--- a/app/views/research_sessions/preview.html.erb
+++ b/app/views/research_sessions/preview.html.erb
@@ -1,11 +1,13 @@
 <%= render "research_sessions/progress", preview: true %>
 
 <form class="print-section screen-only js-print-section" action="#">
-  <select name="area-to-print" id="area-to-print" class="print-section__select js-area-to-print">
-    <option value="#consent-form,#info-sheets">Consent Form and Info Sheet</option>
-    <option value="#consent-form">Consent Form</option>
-    <option value="#info-sheets">Info Sheet</option>
-  </select>
+  <label for="area-to-print" class="selectfield">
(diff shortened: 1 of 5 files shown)


app/views/research_sessions/preview.html.erb
```diff
@@ -1,11 +1,13 @@
...
-    <option value="#consent-form">Consent Form</option>
-    <option value="#info-sheets">Info Sheet</option>
-  </select>
+  <label for="area-to-print" class="selectfield">
+      <select name="area-to-print" id="area-to-print" class="print-section__select js-area-to-print">
+        <option value="#consent-form,#info-sheets">Consent Form and Info Sheet</option>
+        <option value="#consent-form">Consent Form</option>
+        <option value="#info-sheets">Info Sheet</option>
+      </select>
+  </label>
```
rgarner: So does the `label` only exist to give us a hook for `.selectfield::before`?

rgarner: (and why is it a label if we don't have any text content for it? Is it to widen the clickable area somehow?)

joelcarr: Yep, as `select` tags don't allow for pseudo elements. And using a `label` will lets the user interact with the `select` field.

rgarner: Ok, great! So if this is our established pattern for `select` and this is the markup it will generate, we should have a `Barnardos::ActionView::FormBuilder#select` method. It doesn't necessarily have to be done here, we should just acknowledge that This Is How We Do It (perhaps paste the example into the `/gallery` for now?).

PR #126, opened 2017-10-06
wip! swap consent buttons for 'toggle' link

joelcarr:
The buttons on the preview screen were difficult to understand. After speaking with @grockett, we thought to explore the idea of having a toggle instead. Let us know what you think.

before:
image

after:
image
image


grockett:
Can't believe you did that so quick!

I think if we label it up with "yes/no" for the people that might be less
familiar with the toggles from mobile it'll be much tidier and less
confusing than the two buttons 👍

rgarner:
This looks very nice, lots nicer than the previous styling visually, and we should probably put it (or something like it) into our visual language.

I don't think it should be used here though, for two reasons:

* Language: this is generally used to denote a setting that can be on or off. While consent is certainly two state here, I'm not sure it wholly corresponds to an on/off state.
* Behaviour: We are effectively using this as a link, which violates a user's expectation of what should happen when it's clicked. My personal expectations (open to challenge!) of a toggle button that looks like this are: The button animates, and (possibly) something on the page either changes state instantly or indicates instantly (a spinner's pretty usual) that it's about to do so. The way this is currently, no animation happens and a full server roundtrip is triggered. If the latency of that is high, away from our development machines and on a corporate network, this will result in unexpected behaviour ("that switch I clicked didn't immediately change state like I'm used to on the iPad, what's wrong?")†

In summary, style-wise it looks a *lot* nicer, but behaviour-wise I think it's more confusing than the two links we currently have. We could address the behaviour point but I'm not sure we can address the language part – the researcher will have to mentally map the yes/no value they're thinking of to the purely visual on/off state.

† Microsoft are currently doing exactly this roundtrip thing in switching from the Hotmail standard interface to the new beta. In this case, though, the switch obviously changes state before the roundtrip is triggered, although they don't make it completely clear that the roundtrip is happening!

joelcarr:
I definitely agree with you @rgarner. This feature is great for dynamic UIs which we do not have and I can see the confusion. It is somewhat covered up with our local environments due to 0 latency. I do like the idea of a toggle of some sort though as the two buttons were unclear which one was active.

rgarner:
Closing pending [conversations](https://www.youtube.com/watch?v=MZxXj0Z90W4)

PR #127, opened 2017-10-09
Add an unreleased 0.3.0

rgarner:
- Also add the yes/no ordering change


rgarner:
@joelcarr if you can review/merge this before https://github.com/barnardos/consent-form-builder-rails/pull/122 then after rebasing you've got a place to put your select box item :)

PR #136, opened 2017-10-11
Allow additional subfields to show/hide on IE11

joelcarr:
# ["Other" checkboxes do not show/hide associated fields on IE11](https://trello.com/c/vU78qwtO/197-other-checkboxes-do-not-show-hide-associated-fields-on-ie11)

We need to use the lodash polyfill for `includes` script and not the inbuilt browser
emacscript `array.includes`


joelcarr:
giphy-downsized-large

PR #143, opened 2017-10-20
Correct a readme typo

tahb:
* The rest of this file uses `postgres` instead of `postgresql` unless linked so hopefully it's meant to be this way round :)


rgarner:
Received with thanks :)

rgarner:
(Ideally we'd say PostgreSQL everywhere, but this leaves the README in a better state than we left it!)

rgarner:
Closing in favour of #145 due to stalled Circle build here and no way of restarting.

PR #149, opened 2017-10-26
Session naming

rgarner:
# [Session naming](https://trello.com/c/VQbfdtLM/222-8-session-naming)

As a researcher
I want to name my session
So that I get memorable URLs and can make new documents from existing ones*

* **This PR enables that; it doesn't implement it**

This PR will disambiguate session slugs before creating the session by appending a [UUID](https://www.uuidgenerator.net/) to the slug in the event of a name clash, e.g.

- Session 1 has a name of "Bullying in schools" and a slug of `bullying-in-schools`
- Session 2 has a name of "Bullying in schools" and a slug of `bullying-in-schools-32052074-f2a19714-ba4e-11e7-9355-75c69bedbf81`

screen shot 2017-10-26 at 13 09 22

- [x] For the already created sessions we should keep the URLs as they are and name sessions as "1", "2", "220" so that previously saved URLs remain functional
- [x] For new sessions a name should be provided by researcher
- [x] The session name along with a disambiguator if necessary should be used in the URL (e.g. bullying-at-school-{UUID})


app/models/research_session.rb
```diff
@@ -47,4 +53,12 @@ class ResearchSession < ApplicationRecord
   def reached_step?(step)
     Steps.instance.reached_step?(self, step)
   end
+
+  def set_slug_from_name!
+    return if name.blank?
+
+    slug = name.strip.downcase.tr(' ', '-')
```
joelcarr: What if people use punctuation such as "My research project for Barnardo's!"?

joelcarr: Or even an emoji "My research project for Barnardo's! 🤘"

joelcarr: image

Doesn't seem to be a problem!

PR #151, opened 2017-10-30
Change FactoryGirl to FactoryBot

rgarner:
While the etymology of FactoryGirl might have been innocent enough,
let's not pretend we can hire for diversity while knowingly
perpetuating a harmful stereotype. Go gender-neutral, it makes things
better.

And thanks, ThoughtBot, for being ThoughtfulBot.

https://github.com/thoughtbot/factory_bot/blob/master/NAME.md

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,7 +29,7 @@ group :test do
   gem 'cucumber-rails', require: false
   gem 'database_cleaner'
   gem 'launchy'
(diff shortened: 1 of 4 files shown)


joelcarr:
Changelog?

rgarner:
No changelog required – no change in `app/`

PR #153, opened 2017-10-30
Replace stray FactoryGirl reference

rgarner:
Post-merge of two unrelated PRs (session-naming and factory-bot
branches), we've ended up with a build-breaked stray FactoryGirl
reference.

Replacing it here.


joelcarr:
🤖

PR #155, opened 2017-11-01
Create page titles for each step

rgarner:
# [Create page titles for each step](https://trello.com/c/iZY1S1ZC/200-3-create-page-titles-for-each-step)

Make page titles reflect where we currently are in the process.

This also points out that our "preview" stage is a bit convoluted –
a session will be at the "incentive" step. We don't really have a
"finished" or "preview" state. We should, but that's tech debt.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -10,9 +10,9 @@ module ApplicationHelper
 
   def title(research_session, step)
     if research_session.nil?
-      'Consent Form Builder'
+      I18n.t('application.title')
     elsif research_session.new_record?
-      'Create a new form – Consent Form Builder'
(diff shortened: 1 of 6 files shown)


app/helpers/application_helper.rb
```diff
@@ -7,4 +7,16 @@ def link_to_current_sha(sha_getter = -> { `git rev-parse HEAD` })
 
     link_to sha[0..7], COMMIT_STEM + sha
   end
+
+  def title(research_session, step)
+    if research_session.nil?
+      'Consent Form Builder'
```
joelcarr: We should place this copy in `en.yml`

rgarner: Fo'sho

PR #156, opened 2017-11-01
Share save session

joelcarr:
# [Share / Save consent form session](https://trello.com/c/zU3zqiy1/92-5-share-save-consent-form-session)

As a consent form creator,
I would like to be sent a URL of the session by email,
so that I can easily return to the session later

Created a new 'Share' button that lives in a new header nav area. When clicking on the share button, a tooltip opens with content on how to share along with a link to share the URL via email.

image
image
image


app/views/research_sessions/_share.html.erb
```diff
@@ -0,0 +1,27 @@
...
+
+    <p><a href="mailto:?subject=<%= URI.escape(@research_session.name) %>&body=<%= URI.escape(request.original_url) %>" target="_blank">You can email this link by clicking here</a></p>
+    <p>You can add this page to your bookmarks by pressing:</p>
+    <ul class="bullet-point-list">
+      <li>Command/Cmd + D on Mac</li>
+      <li>CTRL + D on Windows</li>
+    </ul>
+  </div>
+  <div class="share__overlay js-share-target js-share-trigger"></div>
+</div>
```
joelcarr: Should probably place all this copy in `en.yml`

app/views/research_sessions/_share.html.erb
```diff
@@ -0,0 +1,27 @@
...
+        type="text"
+        id="share-input"
+        class="share__input"
+        value="<%= request.original_url %>"
+        readonly />
+
+      <button class="button button--no-margin button--copy js-copy-url-button" data-clipboard-target="#share-input">Copy</button>
+    </div>
+
+    <p><a href="mailto:?subject=<%= URI.escape(@research_session.name) %>&body=<%= URI.escape(request.original_url) %>" target="_blank">You can email this link by clicking here</a></p>
```
rgarner: Could we keep the copy here to a call to action "Email this link"? "Click[ing] here" feels like a copy smell.

app/views/research_sessions/_share.html.erb
rgarner: I wouldn't worry about it for now – if it comes to real I18n later, we'd probably [internationalise the whole partial](http://guides.rubyonrails.org/i18n.html#localized-views) rather than breaking things up into copy chunks and finding some YAML home for them.

We are getting other benefits from the I18n of labels, chiefly the centralisation of both label and ActiveRecord validation messages to point to the same thing.

app/views/research_sessions/_share.html.erb
rgarner: (and if we do that, could we also remove the "You can" from the "bookmark" part?)

app/views/research_sessions/_share.html.erb
rgarner: (that said there's also the [safe HTML](http://guides.rubyonrails.org/i18n.html#using-safe-html-translations) option)

PR #158, opened 2017-11-01
Merge Topic / Purpose and update copy

rgarner:
# [Merge Topic / Purpose and update copy](https://trello.com/c/YIQw4isH/208-3-merge-topic-purpose-and-update-copy)

> As a researcher
> I want to provide topic and purpose in a clear and easy to understand way
> So that research participants know what is the aim of my research

screen shot 2017-11-02 at 10 38 51


rgarner:
@Jesskyri @joelcarr @jeddy3 not sure whether we need the text areas to be actually inline/text *fields*

Jesskyri:
@rgarner @joelcarr shall we use this placeholder text? 
*topic* Barnardo’s is doing research to learn about [how young people think of consent and research] *purpose* so that we can [talk to them in a way that they understand.]

rgarner:
@sampling ah yes, placeholder text, thank you for the reminder :)

joelcarr:
@rgarner they should be text fields as the desired outcome shouldn't have any formatting

rgarner:
Thanks @joelcarr, will address at same time as placeholders

app/views/research_sessions/purpose.html.erb
```diff
@@ -3,14 +3,6 @@
   <%= render 'error_summary' %>
```
joelcarr: We should delete the whole file as it is no longer being used.

rgarner: Oh yes, good spot

app/views/research_sessions/preview.html.erb
```diff
@@ -24,34 +24,27 @@
 
 <div class="print-area js-print-area" id="info-sheets">
   <section>
-    <h2 class="heading-medium">About this research project</h2>
-    <p class="super-text">This information sheet will tell you more about the
-    research so that you can decide whether or not <%= you_or_your_child %> would like to take part.</p>
-  </section>
-
-  <section>
```
rgarner: @joelcarr does this look like a mistaken removal to you?

PR #160, opened 2017-11-02
CHORE: Move Questions into their own controller

rgarner:
`ResearchSession` and `Question` (even though the latter concept doesn't have a specific class) are separate concepts that need handling by separate controllers.

Move Questions into their own controller.

This removes the need for us to remember which methods in `ResearchSessionsController` are being used to manage the lifecycle of a `ResearchSession` and which are updating answers to a particular question. It also enables the use of short-form routing via `research_session_<something>_path(@research_session)` and opens up the `create` and `update` methods for a `ResearchSession` should we need them.

It should also make the implementation of https://trello.com/c/qYBbtDC9/87-create-a-new-form-and-info-sheet-from-a-saved-form-and-info-sheet-aka-research-session significantly cleaner.


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * Progress bar should reflect latest design
 * [FEATURE] - [Share / Save consent form session](https://trello.com/c/zU3zqiy1/92-5-share-save-consent-form-session)
 * [BUG] - [Expenses / refreshments improvements](https://trello.com/c/2mT0SkHT/212-2-expenses-refreshments-improvements)
+* [CHORE] - Separate controllers into `ResearchSessionsController` and `QuestionsController`
```
joelcarr: How come you have decided to add a changelog entry?

rgarner: Dangerbot decided for me. Path of least resistance.

joelcarr:
You only #YOLO once

PR #165, opened 2017-11-09
Add yes/no radiobuttons on top of Where&When, Expenses and Incentives

rgarner:
# [Add yes/no radiobuttons on top of Where&When, Expenses and Incentives](https://trello.com/c/Ncnz0D6S/229-2-add-yes-no-radiobuttons-on-top-of-wherewhen-expenses-and-incentives)

> As a researcher
> I want to be able to skip the parts which are optional
> So that I don't waste time on parts which do not apply to my research

Add yes/no radio buttons to each wholly optional step to make it easier to skip over them without reading. Only show the associated fields if "Yes" is selected. Store the radio button's value such that we can occlude fields we've already entered by clicking "no" and have the "no" persisted (but not lose any of the other fields)


db/migrate/20171109124132_add_enable_fields_for_optional_steps.rb
```diff
@@ -0,0 +1,7 @@
+class AddEnableFieldsForOptionalSteps < ActiveRecord::Migration[5.1]
+  def change
+    add_column :research_sessions, :where_when_enabled, :boolean, default: false
+    add_column :research_sessions, :expenses_enabled, :boolean, default: false
```
joelcarr: We should check to see if any of the expenses fields have been filled in to set the field value

rgarner: @joelcarr You mean for each of the steps, for this migration alone? I'll take a look.

app/views/research_sessions/preview.html.erb
```diff
@@ -113,7 +113,7 @@
 
     <p>The data will be kept for <%= edit_link_for(:shared_duration) %>.</p>
   </section>
-  <% if @research_session.when_text.present? or @research_session.duration.present? %>
+  <% if @research_session.where_when_enabled %>
```
rgarner: I'm not sure this is a straight swap. I think we need both, or we need a validator to check that at least one value is `#present?` when `where_when_enabled` is true.

rgarner: (if we carry on like this, we're quite within our rights to say "yes" to enabling this section, then fill nothing in and have blank values in the preview)

db/migrate/20171109124132_add_enable_fields_for_optional_steps.rb
rgarner: Ok, this is rebased in now.

PR #166, opened 2017-11-15
Ignore '£' in currency fields

joelcarr:
# [When a sterling sign is added in, the preview generates 0.00](https://trello.com/c/IHdoFxEl/246-3-bug-when-a-sterling-sign-is-added-in-the-preview-generates-000)

> As a user
> I want to be able to use currency character in money fields
> So that it doesn't affect the saved amount

Add validation to currency fields to ensure numerical values are given.
Strip '£' as it is a common use case.

### DIFF
--- a/app/controllers/research_sessions/questions_controller.rb
+++ b/app/controllers/research_sessions/questions_controller.rb
@@ -42,6 +42,15 @@ module ResearchSessions
       params.require(:research_session).permit(ResearchSession::Steps::PARAMS[step]).tap do |p|
         p['methodologies']&.reject!(&:blank?)
         p['recording_methods']&.reject!(&:blank?)
+
+        %w[
+          travel_expenses_limit
+          food_expenses_limit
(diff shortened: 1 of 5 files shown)


joelcarr:
Hold your horses! CHANGELOG.md

app/controllers/research_sessions/questions_controller.rb
```diff
@@ -42,6 +42,15 @@ def question_params
       params.require(:research_session).permit(ResearchSession::Steps::PARAMS[step]).tap do |p|
         p['methodologies']&.reject!(&:blank?)
         p['recording_methods']&.reject!(&:blank?)
+
+        %w[
+          travel_expenses_limit
+          food_expenses_limit
+          other_expenses_limit
+          incentive_value
+        ].each do |attr|
```
rgarner: I wonder if we ought to factor this into `ResearchSession.currency_params` or if that's a bit premature

PR #172, opened 2017-11-30
Remove additional researcher / Update methodologies

rgarner:
This PR is an amalgamation of two cards. The overall effect is to remove the "Who is doing the research?" section and replace it with "What happens in this research session?" and to update the methodologies to reflect changes in card 2. below. See the following two screenshots for the step question pages (and live preview in the first) and the third and last screenshot for the overall effect on final preview.

## 1. [Remove additional researchers from the builder](https://trello.com/c/sDbgJ6dQ/276-8-remove-additional-researchers-from-the-builder)

image

## 2. [Update the methodologies list in the form builder and info sheet](https://trello.com/c/oI7k0fFy/210-5-update-the-methodologies-list-in-the-form-builder-and-info-sheet)

image

# Together, these result in:

image


Jesskyri:
Hey, for the first one [Remove additional researchers from the builder] is it possible that the researcher's full name is only shown in the first instance of it and not in the three subsequent places on the information sheet? It feels a bit overkill/ formalised

Jesskyri:
For the second one, [Update the methodologies list in the form builder and info sheet] this is my fault but we don't need the brackets around second checklist "An observation of trying out the (new tool) we're designing" it should just be "An observation of trying out the new tool we're designing". Initially when we were doing research, we thought about having somewhere to specify the tool to make it clearer but we're not doing this for this sprint.

rgarner:
Those weren't in the cards – we'll need new cards for these. Yes, it's possible.

PR #173, opened 2017-12-04
Add concerns and complaints

rgarner:
# [Add "Concerns and complaints" to the information sheet (static)](https://trello.com/c/Rvs9PGTM/279-2-add-concerns-and-complaints-to-the-information-sheet-static)

image

### DIFF
--- a/app/views/research_sessions/preview.html.erb
+++ b/app/views/research_sessions/preview.html.erb
@@ -122,7 +122,7 @@
     <dl class="definition-list">
       <div>
         <dt class="definition-list__term">Telephone:</dt>
-        <dd>02084987869</dd>
+        <dd>+442084987869</dd>
       </div>
       <div>
         <dt class="definition-list__term">Email:</dt>
--- a/app/views/research_sessions/preview.html.erb
+++ b/app/views/research_sessions/preview.html.erb
@@ -115,6 +115,23 @@
       confidential.</p>
   </section>
   <section>
+    <h3 class="subtitle-small" id="concerns">Concerns and complaints</h3>
+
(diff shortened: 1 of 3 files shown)


app/views/research_sessions/preview.html.erb
```diff
@@ -114,6 +114,23 @@
...
   </section>
+  <section>
+    <h3 class="subtitle-small" id="private">Concerns and complaints</h3>
+
+    <p>If you have any concerns or a complaint, or would like to have your data deleted please contact:
+
+    <dl class="definition-list">
+      <div>
+        <dt class="definition-list__term">Telephone:</dt>
+        <dd>02084987869</dd>
```
tosbourn: Would it be worth making this a link with `href="tel:+442084987869"` to help folk browsing from outside the UK correctly contact the correct number.

joelcarr: As mentioned offline, this page is for printing so the functionality to call isn't needed here. However prepending the number with +44 should be done.

app/views/research_sessions/preview.html.erb
```diff
@@ -114,6 +114,23 @@
       collected as part of the research is stored securely and kept strictly
       confidential.</p>
   </section>
+  <section>
+    <h3 class="subtitle-small" id="private">Concerns and complaints</h3>
```
joelcarr: Is there a reason for the ID?

rgarner: We meant to link to individual sections, which we aren't currently doing, but to keep with the convention I've updated it to `concerns`

PR #174, opened 2017-12-04
Update recording list

rgarner:
# [Update the recording list in the information sheet and builder](https://trello.com/c/ohPkaN6h/211-3-update-the-recording-list-in-the-information-sheet-and-builder)

> _As a research participant
> I want my Information Sheet to be as simple as possible
> So that I can understand what I am consenting to_

image

### DIFF
--- a/app/helpers/research_sessions_helper.rb
+++ b/app/helpers/research_sessions_helper.rb
@@ -45,6 +45,14 @@ module ResearchSessionsHelper
     end
   end
 
+  def recording_method_lookup(recording_method)
+    if recording_method.to_s == 'other'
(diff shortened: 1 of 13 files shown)


rgarner:
Hang on, the W in "Wheeee" is capitalised here, is that an issue?

joelcarr:
Please go fix your test data then `git push -f`

rgarner:
Yeah, unfortunately we can't tell if a genuine abbreviation would be used at the start that a random `downcase` would nix. I'll leave it; when we get to the live preview it'll make this obvious.

PR #178, opened 2017-12-05
Add 'decide if you want take part' section

joelcarr:
# [Add "Decide if you want the child in your care to take part" part to the info sheet (static)](https://trello.com/c/uzbzifmd/280-2-add-decide-if-you-want-the-child-in-your-care-to-take-part-part-to-the-info-sheet-static)

Add a new section for participants to understand what they / the
person in their custody is taking part in. Create a new class to
have a numbered list.

### DIFF
--- a/app/views/research_sessions/preview.html.erb
+++ b/app/views/research_sessions/preview.html.erb
@@ -165,6 +165,25 @@
     </p>
   </section>
   <section>
+    <h3 class="subtitle-small" id="concerns">
+      Decide if you want<%= 
+        ' the child in your care' if @research_session.unable_to_consent? 
+      %> to take part</h3>
+
+    <p>If you are happy<%=
+      ' for the child in your care' if @research_session.unable_to_consent?
+      %> to take part, we will ask you to sign a consent form. This will ask you to confirm:</p>
+
+    <ol class="numbered-list">
+      <li>that you are happy with the information you have received, and;</li>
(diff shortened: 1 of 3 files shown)


app/views/research_sessions/preview.html.erb
```diff
@@ -136,6 +136,20 @@
     </dl>
   </section>
 
+  <section>
+    <h3 class="subtitle-small" id="concerns">Decide if you want <%= 'the child in your care ' if @research_session.unable_to_consent?  %>to take part</h3>
+
+    <p>If you are happy <%= 'for the child in your care ' if @research_session.unable_to_consent? %>to take part, we will ask you to sign a consent form. This will ask you to confirm:</p>
```
rgarner: Line's a little long here, Mr. @joelcarr. Astonishingly it is also L142, which is the same line @kittysquee had to split earlier. I think we might have to mandate that this file only goes up to 141 lines since L142 is clearly problematic.

tosbourn:
@joelcarr can you rebase to fix the CHANGELOG conflict.

I'm thinking we can probably merge some of these one line additions into the changelog into a longer line explaining some of the text changes.

rgarner:
I'm thinking that also because heavens above and so forth. Let's do that when everything's in `integrate-copy-changes`

joelcarr:
Fixed conflict in `CHANGELOG.md`

Amended the really long line.

PR #179, opened 2017-12-05
Remove "Do I have to take part?" from info sheet

kittysquee:
# [Remove "Do I have to take part" from info sheet](https://trello.com/c/ZJt7BJ6n)

Removes "Do I have to take part" section. 

Changes tests which relied on "Do I have to take part copy" to test consent switch to test different copy.


rgarner:
Hi @kittysquee – since you'll have to rebase anyway because EVERYTHING CONFLICTS ON THE CHANGELOG 🤦‍♂️ 🤦‍♀️ could you fixup the "tests" commit into the other (be good practice for the technique we went over earlier)? Usually we try to make each commit always build (helps with `git bisect`, whenever that's needed).  It should be noted we don't always succeed :wink:

PR #182, opened 2017-12-06
Order preview headers

rgarner:
# [Review the headers/section order of the information sheet](https://trello.com/c/VIAe530f/272-1-review-the-headers-section-order-of-the-information-sheet)

We re-ordered the last bits of the page according to the [document](https://docs.google.com/document/d/1j8HWIUxYyquEZb9RpnZYBVX1SFtwEYg7TqRA3lfO9bI/edit).

We also found that there was no card for:

- Moving preview "storing" section to "confidentiality"
- Remove "Will what I say be kept confidential?"
- Removing extraneous "may" from "About taking part"
- Renaming incorrect "concerns" id to "decide"

Rather than open new cards we've just mopped those up here.

### DIFF
--- a/app/views/research_sessions/preview.html.erb
+++ b/app/views/research_sessions/preview.html.erb
@@ -167,17 +167,6 @@
   </section>
 
   <section>
-    <h3 class="subtitle-small" id="private">Will what I say be kept confidential?</h3>
-    <p>Yes. The goal of design research isnʼt to collect data, itʼs to find the
-      insights we need to improve our services. We do not hand over the recordings; we
-      edit and share insights relevant to the project. Whenever we share these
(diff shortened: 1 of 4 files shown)


app/views/research_sessions/preview.html.erb
```diff
@@ -139,6 +139,30 @@
...
+      Will anyone know what <%= i_or_they %> say in the discussion?
+    </h3>
+
+    <p>
+      The only people who will hear what <%= you_or_your_child %> <%= say_or_says %> in the session
+      will be the researchers running the session, other young people taking part, and
+      service workers facilitating the session.
+    </p>
+
+    <%= edit_link_for(:shared_with) do %>
```
joelcarr: This is outside of any markup

PR #183, opened 2017-12-07
Fix test webpacker compilation

rgarner:
# [Configure RSpec to recompile webpack assets](https://trello.com/c/hLT9twm8/292-configure-rspec-to-recompile-webpack-assets)

Upgrade webpacker to 3.0.2  …

- **Fixes compilation of assets in test environment** (avoids tests running
    against stale React components)
- Also apply the following commands as per webpacker upgrading docs:
    - `yarn upgrade @rails/webpacker --latest`
    - `yarn upgrade webpack-dev-server --latest`

Note that following the upgrade to webpacker 3.x, js-using specs
(preview.html.erb_spec at time of writing) will now be **significantly
(2x) slower if you aren't running webpack-dev-server** when you test (it
looks like webpack-dev-server can manage the state of the compilation,
and in its absence, more compiles are invoked)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ gem 'sass-rails', '~> 5.0'
 gem 'uglifier', '>= 1.3.0'
 gem 'coffee-rails', '~> 4.2'
 gem 'jbuilder', '~> 2.5'
-gem 'webpacker', '~> 2.0'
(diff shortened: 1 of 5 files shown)


joelcarr:
Will approve on green

PR #184, opened 2017-12-07
Live preview topic/purpose

rgarner:
# [Live preview: Topic / Purpose](https://trello.com/c/JiSAAKBv/256-3-live-preview-topic-purpose)

> As a researcher
> I want to see how I am affecting the preview as I edit live values
> So that my changes are obvious and I don't have to skip between steps

Add a live preview to topic. Use the same component on the preview.

## What this also fixes:

We noticed that Jest wasn't being run as part of the main build. Oops! The `ResearcherPreviews/index.test.js` tests were broken for this reason (they referred to old copy). 

1. fix those, and 
1. add Jest to the main build, and
1. fix a problem with `ResearcherPreviews` where `you_or_your_child` wasn't being added
1. fix a problem with `ResearcherPreviews` where the phone # wasn't being rendered

## THIS IS NO LONGER WIP BECAUSE WE FIXED:

- [x] ~~we could still move common code into a base live preview class~~
- [x] ~~we aren't using `component_params` on the live preview side~~

(and don't yet care about)

- [x] the I18n story is a little clunky

## Screenshot:

image

### DIFF
--- a/app/views/research_sessions/questions/topic.html.erb
+++ b/app/views/research_sessions/questions/topic.html.erb
@@ -31,14 +31,12 @@
           <%=
             react_component(
               'TopicPurposePreviews',
-              {
-                topic: @research_session.topic,
-                purpose: @research_session.purpose,
+              component_params(:topic, :purpose).merge(
                 labels: {
                   topic: t('research_session_attr_labels.topic'),
                   purpose: t('research_session_attr_labels.purpose')
                 }
-              }
+              )
             )
           %>
         </section>
--- a/app/views/research_sessions/questions/topic.html.erb
+++ b/app/views/research_sessions/questions/topic.html.erb
@@ -31,7 +31,7 @@
           <%=
             react_component(
               'TopicPurposePreviews',
-              component_params(:topic, :purpose).merge(
+              current_step_params.merge(
                 labels: {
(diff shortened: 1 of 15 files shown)


rgarner:
@sampling is point 3 above the bug you were going to raise? If so we've been forced to deal with it on our way through!

/cc @joelcarr

app/views/research_sessions/questions/topic.html.erb
```diff
@@ -3,23 +3,45 @@
...
+
+      <div class="reactive-container__preview">
+        <section class="reactive-preview__section">
+          <%=
+            react_component(
+              'TopicPurposePreviews',
+              current_step_params.merge(
+                labels: {
+                  topic: t('research_session_attr_labels.topic'),
+                  purpose: t('research_session_attr_labels.purpose')
```
rgarner: This approach – to do translation manually and throw things over the wall – will work here but would not be sustainable for larger volumes of stuff. Suggest a framework such as [i18next](https://github.com/i18next/react-i18next) or Yahoo!'s [react-intl](https://github.com/yahoo/react-intl) along with an adapter for the `en.yml` file to load everything up onec and have the same semantics in both Ruby and JS worlds

PR #188, opened 2017-12-12
WIP! Test ResearcherPreviews constructor name hypothesis

rgarner:
Events are broken in staging and deployment branches, we assume because of our use of `constructor.name` (though we're unsure why dev and test and CI continue to work).

Test this hypothesis with `ResearcherPreviews` and an explicit constructor.


rgarner:
Closing in favour of #189

PR #194, opened 2017-12-23
CHORE: Unvarying params

rgarner:
Since the start of live previews, we've been repeating ourselves across those live previews and the final preview with respect to certain params that didn't vary.

Introduce `ResearchSessionsHelper#unvarying_params` as a home for these such that the call sites for `react_component` no longer have this repetition.

The tests are a little noisier with respect to mocking but hopefully this helps with the cost of making copy changes that would require parameter changes in two places.

## Side effects

This also fixes a minor copy bug where the "storing" step in live preview mode would show you (out of step with the other live previews which would show "your child")

## Merry Christmas! 

🎄 🎅 🎁

### DIFF
--- a/app/helpers/research_sessions_helper.rb
+++ b/app/helpers/research_sessions_helper.rb
@@ -92,6 +92,9 @@ private
           SharedWith::NAME_VALUES.keys.each_with_object({}) do |attr, result|
             result[attr.to_sym] = I18n.t("preview.shared_with.#{attr}", person: you_or_your_child)
           end
+      },
+      recording: {
+        all_recording_methods: RecordingMethods::NAME_VALUES
       }
     }
     unvarying_params[step] || {}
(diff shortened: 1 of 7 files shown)


app/helpers/research_sessions_helper.rb
```diff
@@ -1,10 +1,14 @@
 module ResearchSessionsHelper
   def current_step_params
     component_params(*flat_params(step))
+      .merge(unvarying_params(step))
+      .merge(able_to_consent: false)
```
tosbourn: If we have `able_to_consent?` as a convenience method I think we should use it consistently and remove this hardcoded false.

tosbourn: After looking more at the context in which this is called, I'm going to leave this as false as it is more intent revealing and keeps things consistent (in case able_to_consent was set against the model before going back to a step).

PR #196, opened 2018-01-02
Change start to server in README

kittysquee:
* Change `start` to `server` in Rails server start instructions


tosbourn:
@joelcarr if you're happy with this can you re-review and I will merge in.

PR #199, opened 2018-01-09
DO NOT MERGE

joelcarr:


joelcarr:
Tried making a temporary deployment of `v0.5.0` but it did not work.

PR #214, opened 2018-01-17
Update Gems

kittysquee:
# [Update gems](https://trello.com/c/eI7UjgVW)

* Updates most Gems in Gemfile
* Syntax changes as recommended by updated version of Rubocop
* `pg` not updated as v 1.0.0 isn't compatible with Rails
* `rspec-rails` not updated as other dependencies require
multiple changes
* All updated Gems locked to current versions
* `features/support/action_view.rb` included in `.rubocop.yml` as issue with MixinUsage
* `db/migrate/*` included in `.rubocop.yml` as we were getting Rubocop offences (as HEREDOC uses lower case characters) and it is not recommended practice to edit these files


kittysquee:
This will need to be added to version 0.8.0 in the CHANGELOG.

joelcarr:
Code looks fine to me 🙌 

I am just recreating the preview app so we can make sure everything works in order to approve this PR :)

lib/tasks/npm.rake
```diff
@@ -6,7 +6,7 @@ require 'English'
 namespace :npm do
   package_json = JSON.parse(File.read('package.json'))
 
-  package_json['scripts'].each do |name, _command|
+  package_json['scripts'].each_key do |name, _command|
```
tosbourn: I think we can remove `_command` here, since `each_key` only grabs the keys (not the values)

PR #215, opened 2018-01-17
Update README with PostgreSQL user upgrade

kittysquee:
* This will rectify issues with running rake


kittysquee:
This will need to be added to version 0.8.0 in the CHANGELOG.

PR #218, opened 2018-01-30
Make sure preview can 404

tosbourn:
# [BUG: 404s for research sessions actually 500](https://trello.com/c/S08RNn5r/242-3-bug-404s-for-research-sessions-actually-500)

Preview wasn't 404ing as we would have expected it would. This was because `research_session_controller` was handling 404 in its own unique way. This change removes that logic from the controller so it follows the same logic as the rest of the system.

### DIFF
--- a/spec/controllers/research_sessions_controller_spec.rb
+++ b/spec/controllers/research_sessions_controller_spec.rb
@@ -54,8 +54,8 @@ describe ResearchSessionsController, type: :controller do
         { research_session_id: 'i-dont-exist', research_session: { name: 'I do not exist' } }
       end
 
-      it '404s' do
-        expect(response.status).to eql(404)
+      it 'redirects to 404 page' do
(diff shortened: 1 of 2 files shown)


spec/controllers/research_sessions_controller_spec.rb
```diff
@@ -54,8 +54,8 @@
         { research_session_id: 'i-dont-exist', research_session: { name: 'I do not exist' } }
       end
 
-      it '404s' do
-        expect(response.status).to eql(404)
+      it 'redirects to 404 page' do
+        expect(response.status).to eql(302)
```
joelcarr: Why a 302 and not a 404?

kittysquee: @joelcarr I think we expect the response to equal 302 (URL redirection) because it's doing a redirect to the 404.

PR #251, opened 2018-04-15
Create new frontend architecture

### LINKED ISSUE #246: [2 days] Investigate approaches to align with Design System
The Design System (on other products) are built using a component paradigm. The separation of concerns is based on solving isolated problems, rather than based on technology. The CFB follows the latter approach. As such, Design System components will need to be translated into this other paradigm. This could be an error-prone and long-winded process that could produce undesirable code quality. This spike is to investigate alternatives e.g. https://evilmartians.com/chronicles/evil-front-part-1

joelcarr:
As part of the milestone [Apply Barnardo's design system](https://github.com/barnardos/consent-form-builder/milestone/1), it will be best to incorporate the component based architecture we have been developing to the Consent form builder.

Following the [Modern Front-end in Rails](https://evilmartians.com/chronicles/evil-front-part-1), this PR implements a new component focused frontend architecture. Here are the following features now enabled:

- Using component folder architecture
- Enabled ['live reloading'](https://github.com/webpack/webpack-dev-server#the-result)
- Use the same code quality tools as [design-system](github.com/barnardos/design-system) and [digital-transformation](github.com/barnardos/digital-transformation)
- Able to write components with 'react like' props to build templates

This is quite the investment but I feel this really brings this project inline with our frontend standards therefore reducing the amount of complexity.

Closes #246.


features/create_a_copy.feature
```diff
@@ -3,6 +3,7 @@ Feature: Create a copy
   I want to create a new session from an existing session
   So that I can avoid typing everything from scratch again
 
+@wip
```
tosbourn: Why was this added and can it be removed?

frontend/init/index.css
```diff
@@ -0,0 +1,67 @@
...
+* {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
+}
+
+/* stylelint-enable */
+
+/**
+ * Foundation proportional font-sizing system
```
tosbourn: I found this very hard to read, I wonder if this could be explained more simply.

spec/views/research_sessions/preview.html.erb_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'rails_helper'
 
-describe 'research_sessions/preview' do
+xdescribe 'research_sessions/preview' do
```
tosbourn: Should we just remove this?

app/views/layouts/application.html.erb
```diff
@@ -13,11 +13,10 @@
   <% end %>
   </title>
   <%= csrf_meta_tags %>
-  <%= stylesheet_link_tag 'application', media: 'all' %>
-  <%= stylesheet_link_tag 'print', media: 'print' %>
```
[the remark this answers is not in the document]
joelcarr: We can either create a new pack like `frontend/packs/print.js` which can import style sheets required for print. Or we might find it best to do the media query. It will depend on what requirements we get from our components.

config/environments/production.rb
```diff
@@ -1,4 +1,7 @@
 Rails.application.configure do
+  # Verifies that versions and hashed value of the package contents in the project's package.json
+  config.webpacker.check_yarn_integrity = false
```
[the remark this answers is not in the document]
joelcarr: This came from the webpacker install. I don't see the config in the [Evil Martian's example](https://github.com/progapandist/evil_chat/blob/master/config/environments/production.rb). Do you think we should remove it?

config/webpacker.yml
```diff
@@ -23,16 +29,40 @@ default: &default
...
-    port: 8080
     https: false
+    host: localhost
+    port: 3035
+    public: localhost:3035
+    hmr: false
+    # Inline should be set to true if using HMR
+    inline: true
+    overlay: true
+    compress: true
```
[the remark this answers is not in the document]
joelcarr: In recent frontend architectures you will always compress your assets as your have Source Maps to expand the code.

features/create_a_copy.feature
joelcarr: The preview state will no longer load as it is built from react components which have been moved to start fresh.

frontend/init/index.css
joelcarr: This was copied and pasted straight from the [design-system](https://github.com/barnardos/design-system/blob/master/src/index.css#L32-L41).

spec/views/research_sessions/preview.html.erb_spec.rb
joelcarr: We can enable these tests when the new components for the preview are done. We should not need to change the tests (apart from maybe selectors) which will then validate the new components.

config/environments/production.rb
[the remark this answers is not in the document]
tosbourn: I've just found [some documentation](https://github.com/rails/webpacker/blob/master/README.md#yarn-integrity) on it, looks like it is sensible to keep this.

frontend/init/index.css
tosbourn: Cool, I've opened [an issue](https://github.com/barnardos/design-system/issues/124).

PR #274, opened 2018-04-25
Apply design system to create a new consent form

### LINKED ISSUE #243: Researcher should be able to create a new consent form
## Description

Using the design system, create components to replace the [current 'new' template](https://consent-form-builder-stage.herokuapp.com/) for the consent form builder.

## Acceptance Criteria

- [x] Apply design system
- [x] The form should error if no name for the research session is provided

joelcarr:
Closes #243.


frontend/components/SinglelineTextControl/_index.html.erb
```diff
@@ -14,6 +14,6 @@
     name="<%= id %>"
     id="<%= id %>"
     type="text"
-    size="<%= defined?(size) ? size : "" %>"
+    size="<%= defined?(size) ? size : '' %>"
```
tosbourn: This could be rewritten as `<?= size if defined?(size) %>` which is in keeping with how we've done it on line 9

frontend/components/Form/_index.html.erb
```diff
@@ -0,0 +1,4 @@
+<form class="Form" action="/research-sessions" accept-charset="UTF-8" method="post">
```
tosbourn: We should use `research_sessions_path` instead of hardcoding it.

PR #288, opened 2018-05-17
Save and return

joelcarr:
When a user goes to edit a field from the preview page, when they submit they should be sent back to the preview page.


app/views/research_sessions/questions/methodologies.html.erb
```diff
@@ -30,6 +30,11 @@
       ) %>
     <% end %>
 
-    <%= c('Submit', text: 'Continue') %>
+    <% if params['edit-preview'] == '1' %>
+      <%= hidden_field_tag 'edit-preview', '1' %>
+      <%= c('Submit', text: 'Save and return') %>
```
tosbourn: Is it worth moving this hard-coded text into our i18n yml files?

I don't believe we've been doing that in other PRs so if you feel it is better to keep things identical until we've got the new setup working end-to-end then please ignore this.

kittysquee:
With the repetition in these we could extract it out into a helper method.

PR #293, opened 2018-05-24
Where and when

ritapatel7:
@joelcarr I've added the components, however have realised if single text line components used under another component then there is no space between the components. Will we need to add our own css to this and where do you suggest adding it?


joelcarr:
I see what you mean by the texts fields being closely grouped together. In components such as `Step` we have defined them to use `display: grid;` so we can space their child elements using `grid-gap: 2rem`. [Here is an example](https://github.com/barnardos/consent-form-builder/blob/apply-design-system/frontend/components/Step/index.css). We should apply these styles to the `ChoiceControlledContent` component.

ritapatel7:
@joelcarr styling added. Could you please check and approve?

PR #297, opened 2018-05-29
Align expenses, incentives step to design system and add placeholders

ritapatel7:
Rest of the steps so far... Joel working on fixing label ID issues and also incentives step second set of radio buttons is not working

### DIFF
--- a/app/views/research_sessions/questions/expenses.html.erb
+++ b/app/views/research_sessions/questions/expenses.html.erb
@@ -1,43 +1,67 @@
-<div class="main-content">
-  <%= render 'shared/error_summary' %>
-  <h1 class="page-heading">Expenses</h1>
-  <%= barnardos_form_with model: @research_session, url: wizard_path do |form| %>
+<%= c('Step') do %>
+  <%= c('Title', text: 'Expenses') %>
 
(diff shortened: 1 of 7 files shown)


app/views/research_sessions/questions/expenses.html.erb
```diff
@@ -1,43 +1,89 @@
...
-        <div class="reactive-container__form">
+  <%= c('ChoiceControlledContent',
+      is_active: (@research_session.expenses_enabled.present? && [@research_session.expenses_enabled.to_s].include?('true')),
+      controlled_by: '[id="research_session[expenses_enabled_true]"]') do %>
+        <%= c('SinglelineTextControl',
+          id: 'research_session[travel_expenses_limit]',
+          name: 'travel_expenses_limit',
+          value: @research_session.travel_expenses_limit,
+          label: t('research_session_attr_labels.travel_expenses_limit'),
+          hint: 'e.g. 0.00',
```
kittysquee: Could we use a real world example here? i.e. `5.00`

ritapatel7: i've added this in but do we need anyone else to review what we can use as e.g as the old consent form just uses 0.00 as a placeholder

PR #300, opened 2018-06-01
Create new / copy / share

### LINKED ISSUE #267: Researcher should be able to start a new consent form, share current consent for, or create a copy of the consent form
## Description

Using the design system, create components to replace the 'New', 'Create a copy', and 'Share' commands on the [consent form builder steps](https://consent-form-builder-stage.herokuapp.com/research-sessions/research/questions/researcher).

## Acceptance criteria

- [ ] Researcher should be able to create a new consent form
- [ ] Researcher should be able to create a copy of the current consent form
- [ ] Researcher should be able to share the current consent form

joelcarr:
Closes #267.


frontend/components/PrintAreaCommand/_index.html.erb
```diff
@@ -0,0 +1,18 @@
+<div class="PrintAreaCommand">
```
tosbourn: Since printing is an action, should it ever be a link?

joelcarr: @kelliewellingtons suggested that we should only have one primary action on the page which will be to print and we could offer the printing of a particular section as links.

PR #304, opened 2018-06-06
Consent form bugs

ritapatel7:


app/views/research_sessions/questions/where_when.html.erb
```diff
@@ -28,6 +30,7 @@
   <%= c('ChoiceControlledContent',
       is_active: (@research_session.where_when_enabled.present? && [@research_session.where_when_enabled.to_s].include?('true')),
       controlled_by: '[id="research_session[where_when_enabled_true]"]') do %>
+      <p>Please enter the time and date of the session</p>
```
joelcarr: We should use the Paragraph component here.

PR #305, opened 2018-06-06
Update dependencies to ensure builds run correctly

joelcarr:
It seems builds are struggling to run on Heroku as a dependency returns a 503. Update the dependencies to ensure builds run correctly.


joelcarr:
Builds are still failing. Closing.

PR #307, opened 2018-06-07
Allow livepreview to format currency fields

joelcarr:


ritapatel7:
Tested this and £ showing on Live preview and print preview now.

PR #324, opened 2018-06-13
Step preview layout

### LINKED ISSUE #299: Improve live preview
_As a user
I want to see, when possible, the outcome of my changes alongside my answers
So that I can be confident my answers fit into the context of the final consent form_

## Description

When there is enough space, the preview can be placed alongside the form so that it can be seen more easily by the user.

## Acceptance Criteria

- [ ] The width of the product shouldn't be constrained
- [ ] When there is enough space, the preview should be moved alongside the form
- [ ] Otherwise, it should stay below the form
- [ ] Ensure the line length of text doesn't exceed 95 characters

joelcarr:
Closes #299.

### DIFF
--- a/app/views/research_sessions/questions/methodologies.html.erb
+++ b/app/views/research_sessions/questions/methodologies.html.erb
@@ -3,76 +3,80 @@
 
   <%= c('ErrorSummary', errors: @research_session.errors.to_hash) if @research_session.errors.any? %>
 
-  <%= c('Form', action: wizard_path, patch: true) do %>
-    <%= c('OfManyChoiceControl',
-      id: 'research_session[methodologies][]',
-      name:'methodologies',
-      collection: "research_session[#{:methodologies}][]",
-      choices: Methodologies::CHOICES,
-      choices_selected: @research_session.methodologies,
-      legend: 'How will you be gathering information?',
-      choice_type: 'multi',
-      errored: @research_session.errors['methodologies'].present?,
-      validation: @research_session.errors['methodologies'][0]
-    ) %>
-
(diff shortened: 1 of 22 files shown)


app/views/research_sessions/questions/methodologies.html.erb
```diff
@@ -3,76 +3,80 @@
...
-          url: edit_link_for(:researcher_name)
-        ),
-        c(
-          'Output',
-          text: @research_session.researcher_job_title,
-          field: 'researcher_job_title',
-          url: edit_link_for(:researcher_job_title)
-        )
-      ] %>
+      <%= render partial: 'submit_step' %>
```
kittysquee: This can be written as <%= render 'submit_step' %>

kittysquee:
screen shot 2018-06-13 at 14 17 12

The text at the top seems misaligned and the preview seems misaligned with the input.

app/views/research_sessions/questions/methodologies.html.erb
joelcarr: Could this be part of the Ruby refactor?

kittysquee: Certainly could be 👍

joelcarr:
@kittysquee We could apply negative margin to the top of the preview to try and make the text align. What do you think @kelliewellingtons?

kelliewellingtons:
@kittysquee @joelcarr I'm satisfied with the layout and implementation in this PR. 

Any further work would potentially be made redundant by future tickets in which the content is being overhauled.

PR #325, opened 2018-06-18
Fix logo so it scales proportionally

### LINKED ISSUE #312: Fix the scaling of the logo
The logo should [proportional scale](https://github.com/barnardos/design-system/blob/a2339b10f016f6eb97070173bd753908bda66e5f/src/components/Header/index.css#L20-L24) to be aligned with the Design System, so that it remains relative in size to the body text.

ritapatel7:
Closes #312


kelliewellingtons:
Looks good to me!

PR #327, opened 2018-06-20
Adds start page

### LINKED ISSUE #322: Extract product description and prerequisites into start page
_As a user
I want to know what the product is for and what the prerequisites are
so I won't be blocked midway through the journey_

## Description

Relies on `StartLink` component in DS

## Links and resources

- Copy document:
https://docs.google.com/document/d/1g7ET6R6PlLQVHjgrP8dwM51vttOLnTgIQIkUm0aFsbA/edit?usp=sharing

## Acceptance Criteria

- [ ] Create start page

tosbourn:
Closes #322 by adding a start page with the agreed upon text and components.

screen shot 2018-06-20 at 09 54 56

Uses in progress work from https://github.com/barnardos/design-system/pull/172

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,34 +1,3 @@
-# Files in the config/locales directory are used for internationalization
-# and are automatically loaded by Rails. If you want to use locales other
-# than English, add the necessary files in this directory.
-#
-# To use the locales, use `I18n.t`:
-#
-#     I18n.t 'hello'
-#
-# In views, this is aliased to just `t`:
-#
-#     <%= t('hello') %>
-#
-# To use a different locale, set it with `I18n.locale`:
-#
-#     I18n.locale = :es
-#
-# This would use the information in config/locales/es.yml.
-#
-# The following keys must be escaped otherwise they will not be retrieved by
-# the default I18n backend:
-#
-# true, false, on, off, yes, no
-#
-# Instead, surround them with single quotes.
-#
-# en:
-#   'true': 'foo'
-#
-# To learn more, please read the Rails Internationalization guide
-# available at http://guides.rubyonrails.org/i18n.html.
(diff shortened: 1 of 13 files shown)


config/locales/en.yml
```diff
@@ -380,3 +349,10 @@ en:
           - Name
           - Signature
           - Date
+  start_page:
```
joelcarr: Even though scoping isn't working through our `<%= component(...) %>` tags, we should try to format the `en.yml` to do the following so that scoping could work:

tosbourn: Cool, I will change. I had initially went down that road but because the view file is so minimal in this case it felt like you weren't really saving anything.

tosbourn:
@joelcarr `en.yml` updated.

frontend/components/StartLink/_index.html.erb
```diff
@@ -0,0 +1 @@
+<a href="<%= src %>" class="StartLink"><%= text %></a>
```
jeddy3: Soz... I'm a little late on the review.

The StartLink's text [is always](https://github.com/barnardos/design-system/blob/master/src/components/StartLink/index.js#L8) "Start now".

Shall we pick this up in a follow-up PR?

tosbourn: This site uses internationalisation, so we should be passing this text in, even if it will always be the same text we pass in.

I don't think it would make sense for a component to reach out to the translation library, if we did that then we should rewrite all components to use rails helpers.

For example the above would be re-written as:

``` ruby

jeddy3: > This site uses internationalisation, so we should be passing this text in, even if it will always be the same text we pass in.

Makes sense.

